PR #353, opened 2020-08-19
INTYGFV-13036: Add Certificate Events

arepo2s:


mww-sopra-steria:
Very nice and clean code, both the actual code of course but also the test code! Nice job Anna!

I have two additional thoughts that you could consider.

First, the CertificateEventService. There are a few methods that creates events based on Utkast and on IntygHolder. There are a number of if statements in those that are not obvious. I was wondering if it could be a good idea, from readability, to extract some (or all) into private methods with self-explanatory names that could support the understanding of each. 

Second. Regarding unit tests on services that uses the CertificateEventService. For example, when a draft is signed the CertificateEventService is called to create an event for signing. Could it be a good idea to add validation logic to the unit test to make sure that the CertificateEventService is called with the specific event type? All details doesn't need to be validated, but the behavior is an important one (that an event is created of a certain type) and could be verified through the unit tests. Will act as a good regression test for the future...

mww-sopra-steria:
Another thought that came up this morning. How do we deal with messages imported from Intygstjansten, that belongs to certificates issued outside of Webcert? Have we decided not to add events in those cases and if so, what is the reasoning?

web/src/main/java/se/inera/intyg/webcert/web/event/CertificateEventServiceImpl.java
```diff
@@ -0,0 +1,332 @@
...
+            }
+        }
+        return events;
+    }
+
+    private List<CertificateEvent> createEventsFromArende(String certificateId) {
+        return createEventsFromArende(certificateId, null);
+    }
+
+    public List<CertificateEvent> createEventsFromArende(String certificateId, LocalDateTime afterTimestamp) {
```
mww-sopra-steria, line 250: Should this be public?

mww-sopra-steria (approved):
See comment about visibility on createEventsFromArende(String certificateId, LocalDateTime afterTimestamp) in CertificateEventServiceImpl.

I'm approving the PR. Great job!

web/src/main/java/se/inera/intyg/webcert/web/event/CertificateEventServiceImpl.java
arepo2s: To be able to test it(?)

mww-sopra-steria: OK! Understand! Do you think the test gets too complicated by managing it through getCertificateEvents(..)?  

If you have to test it directly consider making it protected or default (package), so it is not exposed more than necessary.

mww-sopra-steria: Could also add a @VisibleForTesting.

marcus-johansson-2s (approved):

PR #362, opened 2020-08-26
INTYGFV-13221: Added missing PDL logging for messages

marcus-johansson-2s:


mww-sopra-steria (changes requested):
See if you could add unit tests to verify that pdl-logging has been made when expected.

peterssond (approved):

mww-sopra-steria (approved):

PR #378, opened 2020-09-24
INTYGFV-10652: Add events modal for drafts

arepo2s:


web/src/main/java/se/inera/intyg/webcert/web/converter/CertificateEventConverter.java
```diff
@@ -0,0 +1,71 @@
...
+            Optional<ExtendedEventMessage> dataForExtendedMessage = getDataForExtendedMessage(event);
+            dataForExtendedMessage.ifPresent(message -> eventDto.setExtendedMessage(message));
+        }
+        return eventDto;
+    }
+
+    private Optional<ExtendedEventMessage> getDataForExtendedMessage(CertificateEvent event) {
+        IntygContentHolder currentCertificate = intygService.fetchIntygDataForInternalUse(event.getCertificateId(), true);
+        if (currentCertificate != null) {
+            WebcertCertificateRelation parent = currentCertificate.getRelations().getParent();
```
karlssone, line 57: Can getRelations() return null?

web/src/main/java/se/inera/intyg/webcert/web/web/controller/api/dto/CertificateEventDTO.java
```diff
@@ -0,0 +1,130 @@
...
+    private ExtendedEventMessage extendedMessage;
+
+    public CertificateEventDTO(CertificateEvent event) {
+        this.certificateId = event.getCertificateId();
+        this.user = event.getUser();
+        this.eventCode = event.getEventCode();
+        this.timestamp = event.getTimestamp();
+        this.message = event.getMessage();
+    }
+/*
```
karlssone, line 42: Remove if not needed.

web/src/main/java/se/inera/intyg/webcert/web/web/controller/api/IntygApiController.java
```diff
@@ -272,7 +277,12 @@ public Response getEventsForCertificate(@PathParam("certificateId") String certi
             LOG.debug("No events for certificate with id {}", certificateId);
         }
 
-        return Response.ok(eventList).build();
+        List<CertificateEventDTO> eventDTOS = eventList
```
karlssone, line 280: If you are going to change any of my other comments i suggest renaming the variable to eventDTOList.

web/src/main/java/se/inera/intyg/webcert/web/converter/CertificateEventConverter.java
arepo2s: Yes, it can. A question is if it is enough to have the condition earlier so that this only runs if there are events that are assumed to have a relation? You think that I should check this also?

(How much can you assume and how much should you be guarding is not always clear for me, appreciate feedback in this area)

karlssone (approved):

PR #455, opened 2021-01-26
INTYGFV-13679: Added null-check of code.

karlssone:


mww-sopra-steria (approved):
See if you could think of a more descriptive commit-comment when you merge.

PR #459, opened 2021-01-30
INTYGFV-13754: Added new resource link (MARKERA_FRAGA_SOM_HANTERAD) t…

mww-sopra-steria:
…o provide client with information if the user is allowed to set an administrative question as handled.


mww-sopra-steria:
The build-check will fail until we have merged the related infra-pr which included the new constant used by Webcert.

karlssone (approved):

PR #460, opened 2021-02-12
INTYGFV-12569: Created a new endpoint and filter for djupintegration POST requests

RobertDanielsson:
Created a filter which sets character encoding to UTF-8 when posting to the new endpoint "/visa/intyg/defaultutf8/{certId}"


karlssone (approved):

web/src/test/java/se/inera/intyg/webcert/web/web/filter/DefaultCharacterEncodingFilterTest.java
```diff
@@ -0,0 +1,68 @@
...
+    public void testEncodingSet() throws IOException, ServletException {
+        when(httpServletRequest.getCharacterEncoding()).thenReturn("UTF-8");
+
+        testee.doFilterInternal(httpServletRequest, httpServletResponse, filterChain);
+
+        verify(httpServletRequest, times(0)).setCharacterEncoding("UTF-8");
+    }
+
+    @Test
+    public void testUtfStringDoesNotGetCorrupt() throws IOException, ServletException {
```
mww-sopra-steria, line 58: Is this really testing the filter? The filter only sets the encoding if it is missing (which the two upper tests does). This test is just just using the reader the mocks returns and reads the json. Could this test be removed?

mww-sopra-steria (changes requested):
Could we think again about the naming of the endpoint? Sorry to be such a pain ;-) Suggestion is that we discuss it on Mondays dev-meeting.

web/src/test/java/se/inera/intyg/webcert/web/web/filter/DefaultCharacterEncodingFilterTest.java
RobertDanielsson: You're right, I've removed the test. Really good feedback :)

mww-sopra-steria (approved):

PR #471, opened 2021-03-05
INTYGFV-13576: Changed Visa doi button to be displayed only when underlying doi is accessible.

mikeyOne:


mww-sopra-steria:
It hurts a little that we have to add DOI-specific logic here.. but you have done a nice job implementing it and it will execute effectively and only when a DOI certificate is being considered. Well done!

mww-sopra-steria (approved):

PR #478, opened 2021-04-07
INTYGFV-10514: Changes to use statusupdate xml in resultMessage and to avoid excessive namespaces i statusupdate xml, and bugfixes

mikeyOne:


web/src/main/java/se/inera/intyg/webcert/web/service/notification/NotificationRedeliveryStatusUpdateCreatorService.java
```diff
@@ -74,10 +64,10 @@
     /**
      * Creates a {@link CertificateStatusUpdateForCareType} based on the information received in the {@link NotificationRedelivery}.
      */
-    public CertificateStatusUpdateForCareType createCertificateStatusUpdate(NotificationRedelivery redelivery, Handelse event)
-        throws IOException, ModuleException, ModuleNotFoundException, TemporaryException {
+    public String createCertificateStatusUpdate(NotificationRedelivery redelivery, Handelse event)
```
mww-sopra-steria, line 67: Maybe rename the method to something more descriptive. Also update the javadoc.

Also look at the private methods.. with the change to manage it as an xml-string the method´s change purpose and for readability should be renamed.

mww-sopra-steria (changes requested):
Excellent improvement of the implementation! Much simpler now!

I have made some comments surrounding naming of methods and a member. Would be great if you went through it and see if you could give them better descriptive names now when their purpose have changed.

web/src/main/java/se/inera/intyg/webcert/web/service/notification/NotificationRedeliveryStatusUpdateCreatorService.java
mikeyOne: Thanks for valuable input. I agree and will implement the suggestions.

notification-sender/src/main/java/se/inera/intyg/webcert/notification_sender/notifications/services/postprocessing/NotificationResultMessageCreator.java
```diff
@@ -120,9 +122,8 @@ public void addToResultMessage(NotificationResultMessage resultMessage, Certific
 
     private void addRedeliveryMessageToResultMessage(NotificationResultMessage resultMessage,
```
mww-sopra-steria, line 123: This method name still refers to a redelivery message. Would be good if it is renamed as well to something like addStatusUpdateToResultMessage..

web/src/main/java/se/inera/intyg/webcert/web/service/notification/NotificationRedeliveryStatusUpdateCreatorService.java
```diff
@@ -62,64 +55,49 @@
...
+        throws TemporaryException, ModuleNotFoundException, IOException, ModuleException, JAXBException {
+
+        CertificateStatusUpdateForCareType statusUpdate;
+
         if (isDeletedEvent(event)) {
             final var careProvider = hsaOrganizationsService.getVardgivareInfo(event.getVardgivarId());
             final var careUnit = hsaOrganizationsService.getVardenhet(event.getEnhetsId());
             final var personInfo = hsaPersonService.getHsaPersonInfo(event.getCertificateIssuer()).get(0);
-            return certificateStatusUpdateForCareCreator.create(event, careProvider, careUnit, personInfo);
+            statusUpdate =  certificateStatusUpdateForCareCreator.create(event, careProvider, careUnit, personInfo);
```
mww-sopra-steria, line 94: If you can, avoid mutable variables and multiple branches. It requires more to read the code because the reader have to keep track of variables that can be updated and also which branch of a statement was executed.

By just returning here instead of storing in statusUpdate you can both get rid of the statusUpdate variable and the else-clause. 

I believe you had to change this from the original code when dealing with the event time. Sorry for not noticing before.

mww-sopra-steria (approved):
Looks good! Two minor things that I noticed (should have noticed before). I'm approving the PR so no need to review again unless you see necessary.

PR #482, opened 2021-05-06
INTYGFV-13907: Updated api for signature service that accepts more lo…

karlssone:
…a alternatives. Also added success log for signatures with signature service.


web/src/main/java/se/inera/intyg/webcert/web/service/underskrift/dss/DssSignatureService.java
```diff
@@ -264,6 +267,8 @@ private InputDocuments createInputDocuments(SignaturBiljett sb) {
 
         signRequestExtensionType.setSignMessage(createSignMessage(intygsId));
 
+        signRequestExtensionType.setVersion("1.4");
```
mww-sopra-steria, line 270: Suggest to extract "1.4" to a constant with meaningful name.

mww-sopra-steria (approved):

PR #511, opened 2021-06-21
INTYGFV-13859: Extend the facade testability API to include supported certificate types and patients that can be used by the Webcert client Welcome-page.

mww-sopra-steria:


web/src/main/java/se/inera/intyg/webcert/web/web/controller/testability/facade/dto/CertificatePatientsResponseDTO.java
```diff
@@ -0,0 +1,40 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package se.inera.intyg.webcert.web.web.controller.testability.facade.dto;
+
+import java.util.List;
+import se.inera.intyg.common.support.facade.model.Patient;
+
+public class CertificatePatientsResponseDTO {
```
karlssone, line 25: Why don't you use lombok annotation for data and constructor?

karlssone (approved):

web/src/main/java/se/inera/intyg/webcert/web/web/controller/testability/facade/dto/CertificatePatientsResponseDTO.java
mww-sopra-steria: I would like to :-) But currently we are using AutoValue in Webcert-web and as I understands it, it is not possible to use both inside the same "module".

PR #557, opened 2021-09-08
INTYGFV-13991: Added certificate version banner informing of old majo…

mikeyOne:
…r text version.


mww-sopra-steria (changes requested):
Looks good. It would be great if we can have some unit tests that validates that the isLatestMajorTextVersion is correctly set. It defaults to true (which is good) but we need to make sure that it actually sets it correct when it is false.

mikeyOne:
Added tests to verify the property is appropriately updated.

mww-sopra-steria (approved):

PR #558, opened 2021-09-08
INTYGFV-13995: Changed error message returned when vardadmin creates …

mikeyOne:
…db on sekretess patient via CreateDraftCertificate.

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/integration/interactions/createdraftcertificate/BaseCreateDraftCertificateValidator.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/integration/interactions/createdraftcertificate/BaseCreateDraftCertificateValidator.java
@@ -92,23 +92,20 @@ public abstract class BaseCreateDraftCertificateValidator {
             return;
         }
 
-        final var certificateAllowedForSekretess = authoritiesHelper.getIntygstyperAllowedForSekretessmarkering().contains(intygsTyp);
-        if (!certificateAllowedForSekretess) {
-            errors.addError(errorCertificateTypeNotAllowed(sekretessStatus, intygsTyp));
+        if (sekretessStatus == SekretessStatus.UNDEFINED) {
+            errors.addError("Information om skyddade personuppgifter kunde inte hämtas. Intygstypen {0} kan inte utfärdas"
+                + " för patient när uppgift om skyddade personuppgifter ej är tillgänglig.", getCertificateDisplayName(intygsTyp));
             return;
         }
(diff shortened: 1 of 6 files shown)


web/src/main/java/se/inera/intyg/webcert/web/integration/interactions/createdraftcertificate/BaseCreateDraftCertificateValidator.java
```diff
@@ -83,13 +83,32 @@ protected void validatePUServiceResponse(ResultValidator errors, Personnummer pe
...
+        if (personnummer == null) {
+            return;
+        }
+
+        final var sekretessStatus = patientDetailsResolver.getSekretessStatus(personnummer);
+        if (sekretessStatus == SekretessStatus.FALSE) {
+            return;
         }
+
+        final var certificateAllowedForSekretess = authoritiesHelper.getIntygstyperAllowedForSekretessmarkering().contains(intygsTyp);
```
mww-sopra-steria, line 95: Do we have a logical error here?

If we can't retrieve the patient from PU, then it will only give an error if it is not possible to create the certificate type for "sekretess"-patient.

It is good that you check above if it is SekretessStatus.FALSE and just do return.

If it is SekretessStatus.UNKNOWN (typing from memory, so I might have the enum-name wrong), then you should give the error that is produced now on line 110.

If it is SekretessStatus.TRUE, then it should first check if it is possible to create that type on "sekretess"-patient. If not, give the error on line 107. If it is OK, then continue to validateHsaUserMayCreateDraft.

web/src/test/java/se/inera/intyg/webcert/web/integration/interactions/createdraftcertificate/v3/CreateDraftCertificateValidatorImplTest.java
```diff
@@ -287,6 +289,27 @@ public void shouldUseModuleIdInSekretessErrorMessageWhenFailureGettingName() thr
         assertTrue(response.getErrorMessages().get(0).contains(certificateType));
     }
 
+    @Test
+    public void shouldNotAddSecondErrorMessageWhenVardadminAndCertificateNotAllowedForSekretessmarkerad() throws ModuleNotFoundException {
```
mww-sopra-steria, line 293: Probably a good idea to complement with at test when SekretessStatus.UNKNOWN.

mww-sopra-steria (changes requested):
I might have interpreted the code incorrectly, but check if we have a logical error in the code.

mww-sopra-steria (approved):

PR #562, opened 2021-09-14
INTYGFV-13877: Added missing event codes for administrative questions…

mhornfeldt:
… and complements

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificateEventsFacadeServiceImpl.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificateEventsFacadeServiceImpl.java
@@ -190,6 +190,16 @@ public class GetCertificateEventsFacadeServiceImpl implements GetCertificateEven
                 return CertificateEventTypeDTO.EXTENDED;
             case KOPIERATFRAN:
                 return CertificateEventTypeDTO.COPIED_FROM;
+            case NYFRFM:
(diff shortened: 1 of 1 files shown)


mww-sopra-steria (changes requested):
Could you add unit tests for the newly added event types?

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificateEventsFacadeServiceImpl.java
```diff
@@ -190,6 +190,20 @@ private CertificateEventTypeDTO getCertificateEventType(EventCode eventCode) {
...
+                return CertificateEventTypeDTO.INCOMING_MESSAGE_HANDLED;
+            case NYFRFV:
+                return CertificateEventTypeDTO.OUTGOING_MESSAGE;
+            case HANFRFV:
+                return CertificateEventTypeDTO.OUTGOING_MESSAGE_HANDLED;
+            case KOMPLBEGARAN:
+                return CertificateEventTypeDTO.REQUEST_FOR_COMPLEMENT;
+            case PAMINNELSE:
+                return CertificateEventTypeDTO.INCOMING_MESSAGE_REMINDER;
             default:
```
mww-sopra-steria, line 207: Do we need to add Complemented as well? And if so, that should be an event that is decorated with more info, if I'm not mistaken.

mww-sopra-steria (changes requested):

PR #586, opened 2021-11-10
INTYGFV-14134: Added possibility to log errors from client 

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/ErrorLogFacadeServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Override
+    public void log(ErrorLogRequestDTO request) {
+        monitoringService
+            .logClientError(request.getErrorId(), getCertificateId(request.getCertificateId()),
+                request.getErrorCode(), request.getErrorMessage(), request.getStackTrace());
+    }
+
+    private String getCertificateId(String certificateId) {
+        return !certificateId.isEmpty() && !certificateId.isBlank() ? certificateId : "NO_CERTIFICATE_ID";
```
mww-sopra-steria, line 48: Believe you will need to check for null.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/ErrorLogFacadeServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    @Autowired
+    public ErrorLogFacadeServiceImpl(MonitoringLogService monitoringService) {
+        this.monitoringService = monitoringService;
+    }
+
+    @Override
+    public void log(ErrorLogRequestDTO request) {
+        monitoringService
+            .logClientError(request.getErrorId(), getCertificateId(request.getCertificateId()),
+                request.getErrorCode(), request.getErrorMessage(), request.getStackTrace());
```
mww-sopra-steria, line 44: Stacktrace can also be null and should be replaced with something like MISSING_STACKTRACE or NO_STACKTRACE.

web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/dto/ErrorLogRequestDTO.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package se.inera.intyg.webcert.web.web.controller.facade.dto;
+
+public class ErrorLogRequestDTO {
+
+    private String errorMessage;
```
mww-sopra-steria, line 24: We just call it message in the model we defined yesterday.

mww-sopra-steria (changes requested):
Looks good!

Just a few minor comments to address.

As a general comment, I think it looks nicer if you only have one "line feed" between "code sections". Take a look at ErrorLogFacadeServiceImpl as an example.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/ErrorLogFacadeServiceImpl.java
mww-sopra-steria: Or maybe more correct MISSING_STACK_TRACE or NO_STACK_TRACE.

mww-sopra-steria (approved):
Nice!

PR #588, opened 2021-11-11
INTYGFV-14207: Added protected person to user and updated texts

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/UserServiceImpl.java
```diff
@@ -67,6 +67,7 @@ public User getLoggedInUser() {
                     .build()
             )
             .signingMethod(getSigningMethod(webCertUser.getAuthenticationMethod()))
+            .protectedPerson(webCertUser.isSekretessMarkerad())
```
mww-sopra-steria, line 70: Needs unit tests.

mww-sopra-steria (changes requested):

mww-sopra-steria (approved):

PR #597, opened 2021-11-23
INTYGFV-14232: Added display of responsible hosp name.

mikeyOne:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/util/UtkastToCertificateConverterImpl.java
```diff
@@ -106,6 +111,10 @@ private Certificate convertToCertificate(Utkast certificate) {
                 certificateToReturn.getMetadata().getTypeVersion())
         );
 
+        if (webCertUserService.hasAuthenticationContext() && webCertUserService.getUser().getParameters() != null) {
```
mww-sopra-steria, line 114: Just a suggestion for increased code readability. Encapsulate the getResponsibleHospName in a private method where you can include the null checking etc.

mww-sopra-steria (approved):
Approved! But I added a comment to improve readability that you can consider.

PR #601, opened 2021-12-01
INTYGFV-14291: Renew body now supports both FK/AG7804. Small text cha…

RobertDanielsson:
…nge in FK7804 body.

The text toggle in LISJP text is according to this: https://inera.atlassian.net/wiki/spaces/IT/pages/9819062/D1.4.1.+Knappar+i+intygsheader#D1.4.1.Knappariintygsheader-modal1.6

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificatesAvailableFunctionsImpl.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificatesAvailableFunctionsImpl.java
@@ -111,7 +111,8 @@ public class GetCertificatesAvailableFunctionsImpl implements GetCertificatesAva
         + "informationsmängder som är gemensamma för båda intygen automatiskt förifylls.\n"
         + "</p></div>";
 
-    public static final String EVENTUAL_COMPLEMENTARY_REQUEST_WONT_BE_MARKED_READY = "Eventuell kompletteringsbegäran kommer inte att klarmarkeras.";
+    public static final String EVENTUAL_COMPLEMENTARY_REQUEST_WONT_BE_MARKED_READY = "Eventuell kompletteringsbegäran kommer inte "
+        + "att klarmarkeras.";
     public static final String EVENTUAL_COMPLEMENTARY_WILL_BE_MARKED_READY = "Eventuell kompletteringsbegäran kommer att klarmarkeras.";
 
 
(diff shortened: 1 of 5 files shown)


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificatesAvailableFunctionsImpl.java
```diff
@@ -412,6 +401,49 @@ public GetCertificatesAvailableFunctionsImpl(AuthoritiesHelper authoritiesHelper
         return resourceLinks;
     }
 
+    private String getRenewBody(Certificate certificate) {
+        if (isLisjpCertificate(certificate)) {
+            final var complementaryText =
+                isUserAndCertificateFromSameCareUnit(certificate) ? "Eventuell kompletteringsbegäran kommer att klarmarkeras."
```
mww-sopra-steria, line 407: Would be great if the logic deciding the message, for renewal, is covered by unit tests.

mww-sopra-steria (changes requested):

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificatesAvailableFunctionsImpl.java
mikeyOne: The check for same or different care unit seems to be based on **unit or care unit** (whichever is present in valdVardEnhet and unit). Apparently there are some confusion in requirements regarding unit vs care unit, but requirements says the check should be on care unit. So, perhaps have another look at whether the check should be strictly on care unit or not...

If strictly care unit there might be useable code in PR 14151

mww-sopra-steria: Good catch @mikeyOne !

mww-sopra-steria (approved):
Nice!

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/UserServiceImpl.java
```diff
@@ -104,7 +104,7 @@ private String getRole(WebCertUser webCertUser) {
             : webCertUser.getRoleTypeName().split(" ")[0];
     }
 
-    private Vardenhet getLoggedInCareUnit(WebCertUser webCertUser) {
+    public Vardenhet getLoggedInCareUnit(WebCertUser webCertUser) {
```
mww-sopra-steria, line 107: Add Override annotation

mww-sopra-steria (approved):

PR #602, opened 2021-12-02
INTYGFV-14151: Added property careUnit to certificate metadata and us…

mikeyOne:
…er objects.

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/UserServiceImpl.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/UserServiceImpl.java
@@ -63,10 +63,10 @@ public class UserServiceImpl implements UserService {
             .loggedInCareUnit(
                 Unit.builder()
                     .unitName(
-                        loggedInCareUnit != null ? loggedInCareUnit.getNamn() : null
+                        loggedInCareUnit.getNamn()
                     )
                     .unitId(
-                        loggedInCareUnit != null ? loggedInCareUnit.getId() : null
+                        loggedInCareUnit.getId()
                     )
                     .build()
             )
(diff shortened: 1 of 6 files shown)


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/UserServiceImpl.java
```diff
@@ -89,4 +103,31 @@ private String getRole(WebCertUser webCertUser) {
         return webCertUser.getRoleTypeName().equalsIgnoreCase("VARDADMINISTRATOR") ? "Vårdadministratör"
             : webCertUser.getRoleTypeName().split(" ")[0];
     }
+
+    private Vardenhet getLoggedInCareUnit(WebCertUser webCertUser) {
+        final var loggedInCareProviderId = webCertUser.getValdVardgivare().getId();
+        final var loggedInUnitId = webCertUser.getValdVardenhet().getId();
+        final var currentCareProvider = webCertUser.getVardgivare().stream()
+            .filter(careProvider -> careProvider.getId().equals(loggedInCareProviderId)).findFirst().orElse(null);
```
mww-sopra-steria, line 111: As the code is written, it can leave the care unit as null. But if the careunit is null, then the state of the user-object is invalid. 

Probably better to throw an exception instead of returning null, which will remove the need of many nullchecks.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/util/UtkastToCertificateConverterImpl.java
```diff
@@ -134,6 +143,15 @@ private Unit getCareProvider(Utkast certificate) {
             .build();
     }
 
+    private Unit getCareUnit(Utkast certificate) {
+        final var careUnitId = hsatkOrganizationService.getHealthCareUnit(certificate.getEnhetsId()).getHealthCareUnitHsaId();
+        final var careUnit = careUnitId != null ? hsatkOrganizationService.getUnit(careUnitId, "basic") : null;
```
mww-sopra-steria, line 148: What does the magic "basic" do? Maybe a constant?

mww-sopra-steria: Can a certificate be without a unit? Should it be considered an error and throw exception instead?

mww-sopra-steria (changes requested):
A few comments surrounding handling of null.

mww-sopra-steria (approved):

PR #605, opened 2021-12-08
INTYGFV-14280: Doctor name should be set from PU

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/util/IntygToCertificateConverterImpl.java
```diff
@@ -117,10 +122,13 @@ private Certificate convertToCertificate(IntygContentHolder certificate) {
     }
 
     private Staff getIssuedBy(HoSPersonal skapadAv) {
+        final var person = puService.getPerson(Personnummer.createPersonnummer((skapadAv.getPersonId())).get()).getPerson();
```
mww-sopra-steria, line 125: Two things... Firstly it is a HSA-ID and secondly you need to use HSA and not PU. I read the jira and I believe that Test has made a mistake referring to PU instead of HSA.

mww-sopra-steria (changes requested):
Use HSA instead of PU. See comment.

PR #608, opened 2021-12-12
INTYGFV-14280: Doctor name should be set from HSA

mhornfeldt:


mhornfeldt:
I had to switch branch because I got multiple merge conflicts that did not go away after mergin several times

mww-sopra-steria:
Will close this pull request as we figured out that we had the information available already, but didn't use it.

PR #609, opened 2021-12-13
INTYGFV-14288: If patient is protected then body should be sent for print link

mhornfeldt:


RobertDanielsson (approved):

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificatesAvailableFunctionsImpl.java
```diff
@@ -192,26 +192,7 @@ public GetCertificatesAvailableFunctionsImpl(AuthoritiesHelper authoritiesHelper
...
-                ResourceLinkDTO.create(
-                    ResourceLinkTypeDTO.PRINT_CERTIFICATE,
-                    PRINT_NAME,
-                    PRINT_DRAFT_DESCRIPTION,
-                    PRINT_PROTECTED_PERSON_BODY,
-                    true
-                )
-            );
-        }
+        addPrintResourceLink(certificate, resourceLinks);
```
mww-sopra-steria, line 195: Consider having a getMethod that returns a print resourcelink instead of passing in the list and adding it inside the method.

mww-sopra-steria (changes requested):

mww-sopra-steria (approved):

PR #621, opened 2022-01-05
INTYGFV-14292: Added protected user approval resource link

mhornfeldt:


mww-sopra-steria (changes requested):
I'm not sure the right solution is to relate this to a Certificate by using ResourceLinks. 

This is related to the User and whether or not the user has accepted the user agreements relating to being a "protected user".

Look at a solution where the User object has information about this and communicates that way to the client.

PR #629, opened 2022-01-25
INTYGFV-14330: Added translation in facade for fields of validation messages

mhornfeldt:


mww-sopra-steria (approved):
Well done to find a solution to this!

PR #633, opened 2022-02-02
INTYGFV-14404: Changed already existing messages to INFO response.

karlssone:


web/src/main/java/se/inera/intyg/webcert/web/integration/interactions/sendmessagetocare/SendMessageToCareResponderImpl.java
```diff
@@ -51,19 +51,26 @@ public SendMessageToCareResponseType sendMessageToCare(String logicalAddress, Se
             arendeService.processIncomingMessage(ArendeConverter.convert(request));
             result.setResultCode(ResultCodeType.OK);
         } catch (WebCertServiceException e) {
-            result.setResultCode(ResultCodeType.ERROR);
             switch (e.getErrorCode()) {
+                case MESSAGE_ALREADY_EXISTS:
+                    result.setResultCode(ResultCodeType.INFO);
```
karlssone, line 56: Should I set setErrorId?

karlssone: This is the response that will be sent to FK. Is this ok?

karlssone:
Should I make any changes in sendMessageToRecipient?

web/src/main/java/se/inera/intyg/webcert/web/integration/interactions/sendmessagetocare/SendMessageToCareResponderImpl.java
```diff
@@ -51,19 +51,26 @@ public SendMessageToCareResponseType sendMessageToCare(String logicalAddress, Se
...
+                    result.setResultText(e.getMessage());
+                    LOG.info("Could not process incoming message to care. Message already exists. {} : {}", e.getErrorCode(),
```
mww-sopra-steria, line 58: Can we include the message id and maybe even the certificate id, to make it easier to find situations when this happens?

Same for the other logs.

mww-sopra-steria (changes requested):
Would be great with some modifications to the log and the message itself that we returns.

web/src/main/java/se/inera/intyg/webcert/web/integration/interactions/sendmessagetocare/SendMessageToCareResponderImpl.java
mww-sopra-steria: Yes!

mww-sopra-steria (approved):

PR #647, opened 2022-02-21
INTYGFV-14253: Add facade endpoint to get version

perfli:


mhornfeldt (changes requested):
Tests should be added for the controller

mhornfeldt (approved):

PR #652, opened 2022-03-02
INTYGFV-14203: Added system banners to configuration endpoint

mhornfeldt:


web/src/test/java/se/inera/intyg/webcert/web/web/controller/facade/ConfigControllerTest.java
```diff
@@ -65,5 +73,19 @@ void getDynamicLinksReturnsLinks() {
             final var response = (Map<String, DynamicLink>) configController.getDynamicLinks();
             assertTrue(response.containsKey("Test"));
         }
+
+        @Test
+        void getConfigurationReturnsBanners() {
```
mww-sopra-steria, line 78: Would be good with a test that verifies that banners other than Webcert isn't included.

mww-sopra-steria (approved):
Nice, will approve but suggest that you add another test before you merge. See comment in code.

PR #654, opened 2022-03-07
INTYGFV-14509: Updates for changed requirements for missing subscription. Added missing subscription modal.

mikeyOne:


web/src/main/java/se/inera/intyg/webcert/web/service/access/AccessServiceEvaluation.java
```diff
@@ -92,11 +95,12 @@
 
     private AccessServiceEvaluation(WebCertUserService webCertUserService,
         PatientDetailsResolver patientDetailsResolver,
-        UtkastService utkastService, IntygTextsService intygTextsService) {
+        UtkastService utkastService, IntygTextsService intygTextsService, SubscriptionService subscriptionService) {
```
mww-sopra-steria, line 98: How can we avoid dependencies to SubscriptionService? Can it be simplified?

web/src/main/java/se/inera/intyg/webcert/web/service/subscription/SubscriptionServiceImpl.java
```diff
@@ -129,25 +121,22 @@ public void acknowledgeSubscriptionWarning(WebCertUser webCertUser) {
...
-            throw new MissingSubscriptionException(String.format("All care providers for user %s are missing subscription",
-                webCertUser.getHsaId()));
-        }
-    }
-
-    private boolean missingSubscriptionOnAllCareProviders(List<String> allCareProviders,
-        List<String> careProvidersWithoutSubscription) {
-        return allCareProviders.size() == careProvidersWithoutSubscription.size()
-            && allCareProviders.containsAll(careProvidersWithoutSubscription);
+        webCertUser.getSubscriptionInfo().setCareProvidersMissingSubscriptionUnmodifiable(List.copyOf(missingSubscriptions));
```
mww-sopra-steria, line 139: Can we find a different solution to deal with the situation that a user already has been warned about missing subscription for a care provider.

mww-sopra-steria, line 139: Can we find a different solution to deal with the situation that a user already has been warned about missing subscription for a care provider.

mww-sopra-steria (changes requested):
Added two comments from the discussions me and Mikael had about the PR.

mww-sopra-steria (approved):

PR #658, opened 2022-03-10
INTYGFV-14460: Added facade service for getting patient from PU

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/patient/GetPatientFacadeService.java
```diff
@@ -0,0 +1,26 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package se.inera.intyg.webcert.web.service.facade.patient;
+
+import se.inera.intyg.webcert.web.web.controller.facade.dto.PatientResponseDTO;
+
+public interface GetPatientFacadeService {
```
perfli, line 23: Not a biggie but I think we should avoid verbs in class names. Also, it seems redundant to have both `Facade` and `Service` in the name. Perhaps we could rename it to `PatientFacade` or `PatientService`?

web/src/main/java/se/inera/intyg/webcert/web/service/facade/patient/GetPatientFacadeServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+import se.inera.intyg.infra.integration.pu.services.PUService;
+import se.inera.intyg.schemas.contract.InvalidPersonNummerException;
+import se.inera.intyg.schemas.contract.Personnummer;
+import se.inera.intyg.webcert.web.service.monitoring.MonitoringLogService;
+import se.inera.intyg.webcert.web.web.controller.facade.PatientController;
+import se.inera.intyg.webcert.web.web.controller.facade.dto.PatientResponseDTO;
+
+@Service
+public class GetPatientFacadeServiceImpl implements GetPatientFacadeService {
+    private static final Logger LOG = LoggerFactory.getLogger(PatientController.class);
```
perfli, line 19: Should the logger be instantiated with the current class?
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(GetPatientFacadeServiceImpl.class);
```

web/src/main/java/se/inera/intyg/webcert/web/service/facade/patient/GetPatientFacadeServiceImpl.java
```diff
@@ -0,0 +1,82 @@
...
+                .build();
+
+        return PatientResponseDTO.create(patient, personSvar.getStatus());
+    }
+
+    private String getFullName(PersonSvar personSvar) {
+        if(personSvar.getPerson().getMellannamn() == null || personSvar.getPerson().getMellannamn().length() == 0) {
+            return personSvar.getPerson().getFornamn() + " " + personSvar.getPerson().getEfternamn();
+        }
+        return personSvar.getPerson().getFornamn() + " " + personSvar.getPerson().getMellannamn() + " " + personSvar.getPerson().getEfternamn();
```
perfli, line 75: The code would be easier to read/process if we introduce a variable for `personSvar.getPerson()`
```suggestion
        final var person = personSvar.getPerson();
        if (person.getMellannamn() == null || person.getMellannamn().length() == 0) {
            return person.getFornamn() + " " + person.getEfternamn();
        }
        return person.getFornamn() + " " + person.getMellannamn() + " " + person.getEfternamn();
```

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/GetPatientFacadeServiceImplTest.java
```diff
@@ -0,0 +1,193 @@
...
+        doReturn(personSvar).when(puService).getPerson(any());
+    }
+
+    void setupPatient(boolean protectedPerson, boolean testIndicated, boolean deceased) {
+        setupPatient(protectedPerson, testIndicated, deceased, true);
+
+    }
+
+    void setupPatient() {
+        setupPatient(false, false, false, true);
```
perfli, line 75: Not important enough to fix for this code, and perhaps IntelliJ gives a slightly better developer experience, but method signatures like this generally is a bad idea.

perfli (approved):

mww-sopra-steria (changes requested):
Good that you didn't implement any logic inside the controller, but instead included it in the facade-service. 

To follow the same pattern as our other facade-controllers and services, try to move the DTO-creation to the PatientController. I noticed that PersonSvar has a status that you need to deal with. Feel free to ping me if you like to discuss options.

mhornfeldt:
> Good that you didn't implement any logic inside the controller, but instead included it in the facade-service.
> 
> To follow the same pattern as our other facade-controllers and services, try to move the DTO-creation to the PatientController. I noticed that PersonSvar has a status that you need to deal with. Feel free to ping me if you like to discuss options.

Right now I need both the status and the patient from the service so that's why I create the DTO there. Do you mean I should only return the patient? In that case how should I handle errors in the frontend which signals that a patient was not found by PU? There are two error messages which I've implemented, one for NOT_FOUND and one when the API-call failed.

mww-sopra-steria (approved):

PR #661, opened 2022-03-24
INTYGFV-14452: Configure redirect for e-leg login and add IDP urls to facade

perfli:


mww-sopra-steria:
Looks like it is not ready for review just yet. Some unit tests missing and a Todo in UserServiceImpl. But regarding the overall solution, it looks good to me.

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/UserServiceImplTest.java
```diff
@@ -81,15 +88,15 @@ void setUp() {
             .when(user)
             .getNamn();
 
-        doReturn(getNavigableCareProvider())
+        lenient().doReturn(getNavigableCareProvider())
```
mww-sopra-steria, line 91: I think it is better to avoid using lenient and instead restructure the tests in different @Nested classes with separate mocking.

mww-sopra-steria (changes requested):
Nice! Have a couple of suggestions that you could look into.

mww-sopra-steria (approved):

PR #666, opened 2022-04-06
INTYGFV-14585: Change to set a subscriptionInfo object on read-only …

mikeyOne:
…user from rehabstod.


mww-sopra-steria (approved):
Good choice!

PR #681, opened 2022-04-28
INTYGFV-14462: Implemented list for previous certificates of patient

mhornfeldt:


mhornfeldt:
@mww-sopra-steria Do you think I should moved the button types (OPEN_BUTTON, SEARCH_BUTTON, RENEW_BUTTON) to its own type instead of having it as a CertificateListItemValueType? For open and renew it makes sense to have them here because they are values of the table. The search button instead is not. I use the ID's to save/access the button tooltips in a map.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/ResourceLinkListHelperImpl.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    private ResourceLinkDTO getConvertedResourceLink(
+                ActionLink link, String savedUnitId, String certificateType,
+                CertificateListItemStatus status, CertificateRelations relations
+            ) {
+        final var convertedResourceLink = getConvertedResourceLink(link, status);
+        if (convertedResourceLink != null) {
+            return convertedResourceLink;
+        } else if (validateRenew(link, certificateType, relations, getCertificateStatus(status))) {
```
mww-sopra-steria, line 156: Little tricky code to read... but you could remove the else and just do an if. Could make it little more readable.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/config/GetStaffInfoFacadeService.java
```diff
@@ -29,4 +29,6 @@ public interface GetStaffInfoFacadeService {
     String getLoggedInStaffHsaId();
 
     boolean isLoggedInUserDoctor();
+
+    List<String> getUnits();
```
mww-sopra-steria, line 33: Could this be be changed to a little more readable name. You don't understand what it returns until you read the actual implementation.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/config/GetStaffInfoFacadeServiceImpl.java
```diff
@@ -58,6 +60,14 @@ public boolean isLoggedInUserDoctor() {
         return webCertUserService.getUser().isLakare() || webCertUserService.getUser().isPrivatLakare();
     }
 
+    @Override
+    public List<String> getUnits() {
+        WebCertUser webCertUser = webCertUserService.getUser();
```
mww-sopra-steria, line 65: Use final var instead for consistency.

mww-sopra-steria (changes requested):
Well done! Quite easy to read most of the code!

Have just a few comments to look at. Minor things.

Relating your question, I don't have an opinion on the matter. Maybe because I can't see the implications of the different alternatives. Follow your judgement and consider maintainability.

mww-sopra-steria (approved):

PR #688, opened 2022-05-10
INTYGFV-14621: Added service to return statistics for user

mhornfeldt:


mww-sopra-steria (changes requested):
The design has a tight coupling to the frontend design. It included concepts like tabs and also routes in the client. Don't you think it would be better to consider an API where you can get the necessary statistics for the user and let frontend manage what type of components and how to deal with routes?

mww-sopra-steria (approved):

PR #691, opened 2022-05-13
INTYGFV-14453: Choose and change care provider

HelenaEdblad:
There are some remaining questions:

- Should updateSingingService and setFeatureForUser be included? 
- Do we need to check for "abonnemang"? 
- When the user logs in without a care provider set we get response 500 Internal Server Error on /api/session-auth-check/ping. Is this something that we should solve in this task? If so, how can we solve it?


HelenaEdblad:
The remaining questions are now fixed and PR is ready to be reviewed.

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/ChangeUnitImplTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+       doReturn(getUnit())
+               .when(user)
+               .getValdVardenhet();
+
+       doReturn(true)
+               .when(user)
+               .changeValdVardenhet(anyString());
+
+        /*doReturn(true)
```
mww-sopra-steria, line 77: Remove commented code.

mww-sopra-steria (approved):
Looks good! Just remove commented code before merging.

mww-sopra-steria (approved):

PR #714, opened 2022-06-09
INTYGFV-14453: Add resource link for choose unit

HelenaEdblad:


web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/GetUserResourceLinksImplTest.java
```diff
@@ -143,6 +146,39 @@ void shallNotIncludeSignedCertificatesListIfUserIsNotDoctor() {
...
+
+        @Test
+        void shallNotIncludeChooseUnitIfOriginIsDjupintegration() {
+            final var user = getUserWithOrigin("DJUPINTEGRATION");
+            final var actualLinks = getUserResourceLinks.get(user);
+            ResourceLinkFacadeTestHelper.assertExclude(actualLinks, ResourceLinkTypeDTO.CHOOSE_UNIT);
+        }
+
+        @Test
+        void shallIncludeChooseUnitIfOriginIsNormalAndHasLoggedInUnit() {
```
mhornfeldt, line 165: If I'm not wrong the user should get the resource link if she has not a set unit. If that is the case then this test should fail because if the user has a set unit the resource link should not be included. The other two tests are not useful since if the user has a set unit the resource link is gonna be excluded no matter what origin. You should do tests checking that a user with no unit doesn't have the resource link if the wrong origin, that the a user with the correct origin and no unit has the resource link, and that a user with the correct origin and a set user doesn't have the resource link.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/GetUserResourceLinksImpl.java
```diff
@@ -45,50 +44,67 @@ private ArrayList<ResourceLinkDTO> getAvailableFunctionsForUser(WebCertUser user
...
+                )
             );
         }
+
         return resourceLinks;
     }
 
+    private boolean isChooseUnitAvailable(WebCertUser user) {
+        final var loggedInUnit =  user.getValdVardenhet();
+        return isOriginNormal(user.getOrigin()) && loggedInUnit != null;
```
mhornfeldt, line 105: Isn't this the opposite logic of what should be implemented? If the logged in unit is not null then it means that the user has chosen a unit and we don't need to display the choose unit modal. Also, for change unit you included the origins NORMAL and UTHOPP. Here you only include NORMAL. Is this correct or should UTHOPP be included?

mhornfeldt (changes requested):

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/GetUserResourceLinksImpl.java
HelenaEdblad: This is correct, when you change unit both NORMAL and UTHOPP is included, but when you choose unit only NORMAL is included.

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/GetUserResourceLinksImplTest.java
```diff
@@ -143,6 +146,39 @@ void shallNotIncludeSignedCertificatesListIfUserIsNotDoctor() {
...
+            doReturn(getUnit())
+                .when(user)
+                .getValdVardenhet();
+
+            final var actualLinks = getUserResourceLinks.get(user);
+            ResourceLinkFacadeTestHelper.assertExclude(actualLinks, ResourceLinkTypeDTO.CHOOSE_UNIT);
+        }
+
+        @Test
+        void shallNotIncludeChooseUnitIfOriginIsDjupintegrationAnHasNoLoggedInUnit() {
```
mhornfeldt, line 163: Test is missing for uthopp

mhornfeldt (approved):

PR #715, opened 2022-06-10
INTYGFV-14454 (fixed history): Change unit

HelenaEdblad:


mhornfeldt (approved):
PR title has wrong format

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/GetUserResourceLinksImpl.java
```diff
@@ -97,6 +97,17 @@ private ArrayList<ResourceLinkDTO> getAvailableFunctionsForUser(WebCertUser user
             );
         }
 
+        if (isChangeUnitAvailable(user)) {
+            resourceLinks.add(
+                ResourceLinkDTO.create(
+                    ResourceLinkTypeDTO.CHANGE_UNIT,
+                    "Byt vårdenhet",
+                    "",
```
mhornfeldt, line 105: Make sure that there is no text from the modal that can be added in either description or body attribute. The more text we can avoid adding in the frontend the better.

PR #730, opened 2022-06-23
INTYGFV-14614: And logic for alternate patient id and reserve id

perfli:
This implementation more closely matches the one in the old AngularJS client:

Person id is changed if
* beforeAlternateSSN is set AND alternateSSN is set AND beforeAlternateSSN != alternateSSN AND alternateSSN is a personId/coordinationId
      OR
* beforeAlternateSSN is NOT set AND personId is set AND personId != alternateSSN AND alternateSSN is a personId/coordinationId

Has reserve id if

 * beforeAlternateSSN is NOT set AND personId is set AND personId != alternateSSN AND alternateSSN is NOT a personId/coordinationId


mhornfeldt (changes requested):
Variables for changed and updated should be merged

mhornfeldt (approved):

PR #733, opened 2022-06-30
INTYGFV-14706: Change to sort care providers, care units and units st…

mikeyOne:
…rictly by alphabet in user object.


mhornfeldt (approved):
Should unit test be added?

mhornfeldt (approved):

PR #741, opened 2022-07-12
INTYGFV-14638: Header link to portal site for private practitioners

SamuelTilly:


web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/ConfigController.java
```diff
@@ -78,7 +81,7 @@ public Response getConfiguration() {
             .filter((banner -> banner.getApplication() == Application.WEBCERT))
             .collect(Collectors.toList());
 
-        return Response.ok(new ConfigurationDTO(version, banners, sakerhetstjanstIdpUrl, cgiFunktionstjansterIdpUrl)).build();
+        return Response.ok(new ConfigurationDTO(version, banners, ppHost, sakerhetstjanstIdpUrl, cgiFunktionstjansterIdpUrl)).build();
```
mww-sopra-steria, line 84: Could you add a test to verify that the ppHost is added to the response correctly? A simple test that can will be a useful regression test to make sure no one mixes up the strings...

mww-sopra-steria (changes requested):
Looks good! See comment about an additional unit test.

mww-sopra-steria (approved):

PR #754, opened 2022-08-09
INTYGFV-14455: Added resource link SUBSCRIPTION_WARNING when user is …

mww-sopra-steria:
…logged in on a care provider that is missing subscription and hasn't already viewed the warning within the current session.


mhornfeldt (dismissed):

mww-sopra-steria:
@mhornfeldt I've added some additional logic that would be great if you could review.

mhornfeldt (approved):

PR #766, opened 2022-09-01
INTYGFV-14631: Change tooltip description for sign and send

HelenaEdblad:


mww-sopra-steria (approved):
The implementation will have to change later when we introduce another certificate type that has the 'sign and send' resource link.

PR #769, opened 2022-09-27
INTYGFV-14370: Updated gradle and dependency versions. Applied necessary code changes. Some cleanup.

mikeyOne:


mww-sopra-steria (approved):
Just to remind us! :-) We have to merge Intygsjanst before Webcert as we have updated the backing service.

PR #772, opened 2022-10-03
INTYGFV-14695: Add launchId as Integration parameter, and cache in Re…

mww-sopra-steria:
…dis, after successful authentication from an Idp.


ChrilleSigerhed (approved):
Looks great!

PR #775, opened 2022-10-07
INTYGFV-14837: Change default value for signed by dropdown

IamFriberg:
Added default values for sign to show all if it is an administrator. Also removed administrator as an option in the same dropdown.


SamuelTilly (approved):
Looks and works good

PR #784, opened 2022-10-28
INTYGFV-14815: Updated failing tests.

IamFriberg:
Updated failing tests.


ChrilleSigerhed (approved):
Nice, now the test are returning the same AccessResultCode as the other sign methods instead of the genererated NO_PROBLEM!

PR #790, opened 2022-11-04
INTYGFV-14941: Use patient address from certificate, instead of PU, if address is missing in PU.

mww-sopra-steria:


ChrilleSigerhed (approved):
Looks great!

PR #792, opened 2022-11-04
INTYGFV-14931: Refactored CertificateSignConfirmationFunctionImpl to only return resourcelink if a draft also exist of the certificate.

ChrilleSigerhed:


mww-sopra-steria (approved):
Nice! Great with a new test that validates it!

PR #803, opened 2022-11-16
INTYGFV-14902: Added link MISSING_RELATED_CERTIFICATE_CONFIRMATION to…

mww-sopra-steria:
… be used when no related certificate can be found. Only used currently for db/doi.


IamFriberg (approved):
Nice!

PR #807, opened 2022-11-22
INTYGFV-14869: Added link SHOW_RELATED_CERTIFICATE to be used to open doi from a db certificate when both exists on the same patient. Also added new api-endpoint to retrieve the certificateId of the related certificate i.e. in this case doi.

mww-sopra-steria:


IamFriberg (approved):
Nice!

PR #808, opened 2022-11-22
INTYGFV-14878: Added messages in patient view for DOI when draft or certificate already exists. Added tests to verify the behaviour.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/CertificateTypeMessageServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+                return getOptionalStringDraftDb(previousIntyg);
+            } else {
+                return getOptionalStringDraftDoi(previousIntyg);
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private static Optional<String> getOptionalStringCertificateDoi(final PreviousIntyg previousIntyg) {
```
mww-sopra-steria, line 86: I think it can be a bit redundant to have the return type in the filename. Maybe could consider att different name.

Same for the other methods.

mww-sopra-steria, line 86: For consistency you could remove the static-modifier. No need to have the methods as static, but it doesn't hur either.

mww-sopra-steria (changes requested):
Very nice! Please have a look at the very minor comments.

mww-sopra-steria (approved):
Great!

PR #815, opened 2022-12-08
INTYGFV-15023: Added interface that handles resource links for SEND_CERTIFICATE.

ChrilleSigerhed:
Implementation handles FK7801 and FK7804 at the moment.


mww-sopra-steria (approved):
Nice! Great to see that logic getting moved to a separate class!

PR #818, opened 2022-12-12
INTYGFV-14909: Added optional attribute title to ResourceLinkDTO.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/dto/ResourceLinkDTO.java
```diff
@@ -103,7 +124,8 @@ public boolean equals(Object o) {
             && Objects.equals(type, that.type)
             && Objects.equals(name, that.name)
             && Objects.equals(description, that.description)
-            && Objects.equals(body, that.body);
+            && Objects.equals(body, that.body)
+            && Objects.equals(title, that.title);
     }
 
     @Override
```
mww-sopra-steria, line 131: Nice catch on updating equals and toString. Could you also update hashCode?

mww-sopra-steria (approved):
Nice work! Approved, but check the comment before merging :-)

web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/dto/ResourceLinkDTO.java
ChrilleSigerhed: Ops, missed that one! :-)

Done!

PR #822, opened 2022-12-20
INTYGFV-15137: Add recipient as sentTo in CertificateMetaData when converting from internal/transport-models to certificate

mww-sopra-steria:


ChrilleSigerhed (approved):
Great job!

PR #830, opened 2023-01-11
INTYGFV-15156: Implementation of logic that adds resourcelink SEND_CERTIFICATE for FK7800.

ChrilleSigerhed:


mikeyOne (approved):
As a reflection, this should probalbly be don for luae-fs as well

PR #831, opened 2023-01-12
INTYGFV-15157: Implementation of Luse (FK7800) in testability API. Added integrationtests for Luse (FK7800).

ChrilleSigerhed:


mikeyOne (approved):
Looks good!

PR #834, opened 2023-01-14
INTYGFV-15134: Disable resource link COPY_CERTIFICATE, and display tooltip explaining why, on locked db/doi for a patient which already has a certificate of the type.

mww-sopra-steria:


mikeyOne (approved):
Looks good

PR #836, opened 2023-01-18
INTYGFV-15135: Refactored CertificateTypeMessageService to include what type of message it returns. Use the messageType for CopyCertificateFunction to include correct resource link body when copying locked draft and there is a draft on a different care provider. This is only relevant for db & doi.

mww-sopra-steria:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/CertificateMessageType.java
```diff
@@ -0,0 +1,25 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package se.inera.intyg.webcert.web.service.facade.impl;
+
+public enum CertificateMessageType {
+    CERTIFICATE_ON_SAME_CARE_UNIT, CERTIFICATE_ON_SAME_DIFFERENT_CARE_UNIT, DRAFT_ON_SAME_CARE_UNIT, DRAFT_ON_SAME_DIFFERENT_CARE_UNIT,
```
ChrilleSigerhed, line 23: Should we remove the _SAME_ section from these enums?

- CERTIFICATE_ON_SAME_DIFFERENT_CARE_UNIT
- DRAFT_ON_SAME_DIFFERENT_CARE_UNIT

ChrilleSigerhed (approved):
Awsome, great work! 

Only got a small remark on the naming of two of the enums in CertificateMessageType, consider changing them before merging! :-)

PR #839, opened 2023-01-20
INTYGFV-15135: Fixed doi scenario for copying locked draft when a certificate exists on a different care provider.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #841, opened 2023-01-26
INTYGFV-15223: Added correct status when draft has been revoked

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/CertificateListItemConverterImpl.java
```diff
@@ -159,7 +159,7 @@ private CertificateListItemStatus getCertificateListItemStatus(String status, Re
             return CertificateListItemStatus.SENT;
         }
 
-        if (status.equals("CANCELLED")) {
+        if (status.equals("CANCELLED") || status.equals("DRAFT_LOCKED_CANCELLED")) {
```
ChrilleSigerhed, line 162: Consider changing hard coded string values to private static final Strings. Just so that we dont have checkstyle errors!

ChrilleSigerhed (approved):
Awsome, left a comment regarding checkstyle errors! Consider changing it before merging if it makes sense to you! :-)

PR #843, opened 2023-01-30
INTYGFV-15226: Refactored creating filled in test certificates from t…

mww-sopra-steria:
…he testability api. Each certificate type implementation has the responsibility to fill in the certificate of a certain type and major version.


ChrilleSigerhed (approved):
Awsome, well done! Looks great!

PR #848, opened 2023-02-01
INTYGFV-15180: Event available for citizen should not be added for db…

mhornfeldt:
… and doi


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificateEventsFacadeServiceImpl.java
```diff
@@ -145,7 +153,11 @@ private CertificateStatus getRelatedCertificateStatus(WebcertCertificateRelation
         }
     }
 
-    private void addAvailableForPatientIfNeeded(List<CertificateEventDTO> events) {
+    private void addAvailableForPatientIfNeeded(List<CertificateEventDTO> events, String certificateType) {
+        if (certificateType.equals("doi") || certificateType.equals("db")) {
```
mww-sopra-steria, line 157: Use constants in Doi/DbModuleEntryPoint instead of hardcoding.

mww-sopra-steria (approved):
See comment before merging! Would be good to use constants instead.

PR #857, opened 2023-02-08
INTYGFV-13856: Activated CSRF-protection for api/certificate and api/question which are only used by the new React-client. Introduced support for the IntegrationTest to manage CSRF cookie and header.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice, well done. 

Exending BaseFacadeIT in the related test classes was a really clean solution! 

Regarding the endpoint we apply CSRF on, why are some endpoints, like /configuration, /fmb, /icf ignored? :-)

mww-sopra-steria:
> Nice, well done.
> 
> Exending BaseFacadeIT in the related test classes was a really clean solution!
> 
> Regarding the endpoint we apply CSRF on, why are some endpoints, like /configuration, /fmb, /icf ignored? :-)

They don't modify any data, so I only added it where it was necessary.

PR #860, opened 2023-02-13
INTYGFV-14827: Implementation of CertificateForPatientService that utilizes…

ChrilleSigerhed:
… redis caching to speed up the search & sorting of the patients certificates.


mww-sopra-steria (approved):
@ChrilleSigerhed Please review the changes made and if you have any concerns, please raise them. If everything looks good I think you can go ahead and merge.

Main notes on the changes:
* Refactored for testability.
* Simplified the implementation.
* Dealt with a few corner cases that we have to consider because certificates are stored both in IT and WC.
* Introduced a more efficient method in IntygService that doesn't make calls to WC´s repository.

PR #871, opened 2023-03-06
INTYGFV-15395: Added SRS resource link

mhornfeldt:


mww-sopra-steria (changes requested):
Looks good!

I think we just have to add another rule to when the SRS feature should be available. SRS is today only available when the Certificate is in Draft-state (this will probably change later with the new Rehabstod <-> Webcert integration, but for now is the correct behaviour).

mww-sopra-steria (approved):

PR #873, opened 2023-03-14
INTYGFV-15424: Added relations to ListIntygEntry that gets retrieved …

ChrilleSigerhed:
…from WC, so that we display correct status in our list of certificates. This is already done for certificates fetched from IT.

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/previous/CertificateForPatientServiceImpl.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/previous/CertificateForPatientServiceImpl.java
@@ -45,7 +45,6 @@ import se.inera.intyg.webcert.web.service.intyg.IntygService;
 import se.inera.intyg.webcert.web.service.user.WebCertUserService;
 import se.inera.intyg.webcert.web.service.utkast.UtkastService;
 import se.inera.intyg.webcert.web.web.controller.api.dto.ListIntygEntry;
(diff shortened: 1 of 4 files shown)


web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/previous/CertificateForPatientServiceImpl.java
```diff
@@ -71,9 +74,16 @@ private List<ListIntygEntry> getCertificatesFromWC(Personnummer patientId, List<
         LOG.debug("UtkastService returned {} certificates", certificatesFromWC.size());
         return certificatesFromWC.stream()
             .map(IntygDraftsConverter::convertUtkastToListIntygEntry)
+            .map(this::setRelations)
             .collect(Collectors.toList());
     }
 
+    private ListIntygEntry setRelations(ListIntygEntry entry) {
+        final var relations = certificateRelationService.getRelations(entry.getIntygId());
```
ChrilleSigerhed, line 82: Ideally this would get done in IntygDraftsConvert, but thats not possible since the method there is static and hinders DI.

mww-sopra-steria: This could possibly have a negative impact on performance. I would like to analyse this a bit deeper before I can finalize a review.

mww-sopra-steria (approved):

PR #878, opened 2023-03-22
INTYGFV-15421: Added uriForCertificateWithSignError with parameter error.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/util/ReactUriFactory.java
```diff
@@ -43,4 +47,13 @@ public URI uriForCertificate(UriInfo uriInfo, String certificateId) {
             .path(urlReactTemplate)
             .buildFromMap(urlParams);
     }
+
+    public URI uriForCertificateWithSignError(UriInfo uriInfo, String certificateId, SignaturStatus signStatus) {
```
mww-sopra-steria, line 51: Maybe a test for this :-)

mww-sopra-steria (changes requested):
Nice! See comment for extra tests :-)

mww-sopra-steria (approved):

PR #880, opened 2023-04-05
INTYGFV-15502: Handle situation when deep-integration parameters includes a reserve-id as alternatePatientSsn. 

mww-sopra-steria:


ChrilleSigerhed (approved):
Looks great, well done!

PR #895, opened 2023-06-29
INTYGFV-15740: Added contact info for questions

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/question/impl/ArendeToQuestionFacadeServiceImpl.java
```diff
@@ -50,6 +51,7 @@
 import se.inera.intyg.webcert.web.service.facade.question.util.QuestionConverter;
 
 @Service(value = "ArendeToQuestionFacadeService")
+@Transactional
```
ChrilleSigerhed, line 54: Im curious to why the transactional annotation had to be added for this service?

mhornfeldt: I got an error otherwise. 

MicrosoftTeams-image (6)

mww-sopra-steria: I think you could change kontaktinfo to EAGER. Same as for komplettering. Then we don't need to add the @Transactional annotation.

mww-sopra-steria (changes requested):
See comment

web/src/main/java/se/inera/intyg/webcert/web/service/facade/question/impl/ArendeToQuestionFacadeServiceImpl.java
mhornfeldt: If I add the same as for komplettering I get this: Unable to build Hibernate SessionFactory; nested exception is org.hibernate.loader.MultipleBagFetchException: cannot simultaneously fetch multiple bags: [se.inera.intyg.webcert.persistence.arende.model.Arende.komplettering, se.inera.intyg.webcert.persistence.arende.model.Arende.kontaktInfo]

mww-sopra-steria: It seems to be a behaviour of Hibernate... They way it handles EAGER. There are solutions/work-arounds and if you want to avoid the @Transactional I would suggest you also add this:

`@Fetch(value = FetchMode.SUBSELECT)`

So it looks something like this in the end:

image

References:
https://hibernate.atlassian.net/browse/HHH-1718
https://stackoverflow.com/questions/4334970/hibernate-throws-multiplebagfetchexception-cannot-simultaneously-fetch-multipl

mww-sopra-steria (approved):
Approved and if you like the suggested solution, please go ahead with it.

PR #897, opened 2023-07-19
INTYGFV-15776: Added new controller to handle launches of webcert from other origins that decorates user with launchFromOrigin attribute.

ChrilleSigerhed:
…quests from rehabstod, and not display error message every time they navigate from rs to wc.


mhornfeldt (approved):

mww-sopra-steria (changes requested):
Instead of redirecting with rs-query param, store "fromRehabstod" in WebcertUser so the information is persisted for the session.

Also as discussed, move to a separate controller.

mww-sopra-steria (approved):
Very nice! Well done!

PR #898, opened 2023-07-20
INTYGFV-15649: Changed default login to siths-wc2 in IdpSelectionFilter.

ChrilleSigerhed:


mhornfeldt (approved):

mww-sopra-steria (changes requested):
Check if /maillink url also need to be handled with wc2 sp metadata.

ChrilleSigerhed:
> 

After looking into this I found that the maillink uses the following urlBaseTemplate url:
```/webcert/web/user/basic-certificate```

This implies that the changes provided also will affect users that launches webcert from a mail link.

mww-sopra-steria (approved):

mww-sopra-steria:
@ChrilleSigerhed Consider adding unit test for the added behaviour and changed metadata.

PR #902, opened 2023-08-06
INTYGFV-14950: Use SpringSessionBackendSessionRegistry instead of SessionRegistryImpl as Redis is used for session repository. This fixes a memory leak where sessions otherwise are stored in SessionRegistryImpl and (almost) never removed.

mww-sopra-steria:


web/src/main/java/se/inera/intyg/webcert/web/auth/WebcertLoggingSessionRegistryImpl.java
```diff
@@ -31,53 +32,69 @@
...
+        final var principal = authenticator != null ? authenticator.getAuthentication().getPrincipal() : null;
+        return isWebcertUser(principal) ? (WebCertUser) principal : null;
+    }
+
+    private boolean isWebcertUser(Object principal) {
+        return principal instanceof WebCertUser;
+    }
+
+    private static String getUserRole(WebCertUser user) {
+        return user.getRoles() != null && user.getRoles().size() == 1 ? user.getRoles().keySet().iterator().next() : "noRole?";
```
mhornfeldt, line 98: Shouldn't it be size >= 1?

mhornfeldt (approved):

web/src/main/java/se/inera/intyg/webcert/web/auth/WebcertLoggingSessionRegistryImpl.java
mww-sopra-steria: It is logic that isn't touched in this PR (other than some refactoring to a private-method). But it seems like it is implemented as intended. If the user has "one" user-role, then it should be part of the loggpost. 

BaseUserDetailsService.decorateIntygUserWithRoleAndAuthorities sets the role, which is always only one.

PR #907, opened 2023-08-14
INTYGFV-15776: Change the users careunit, to the same as the certificate, when launching Webcert with /launch/certificate/{certificateId}.

mww-sopra-steria:


ChrilleSigerhed (approved):
Well done!

PR #908, opened 2023-08-14
INTYGFV-15801: Updated how question draft is created

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/arende/ArendeDraftServiceImpl.java
```diff
@@ -78,9 +71,16 @@ public ArendeDraft getQuestionDraftById(long id) {
...
+    public ArendeDraft create(String certificateId, String subject, String message, String questionId) {
+        ArendeDraft draft = arendeDraftRepository.findByIntygIdAndQuestionId(certificateId, questionId);
+        if (draft != null) {
+            draft.setText(message);
+            draft.setAmne(subject);
+        } else {
+            draft = createDraft(certificateId, questionId, message, subject);
+        }
+        arendeDraftRepository.save(draft);
+        return draft;
```
mww-sopra-steria, line 83: Return the draft that arendeDraftRepository.save(draft) returns. The returned draft must for example contain the id.

web/src/test/java/se/inera/intyg/webcert/web/service/arende/ArendeDraftServiceImplTest.java
```diff
@@ -90,6 +90,51 @@ public void testSaveDraftUpdate() {
         verifyNoMoreInteractions(repo);
     }
 
+    @Test
+    public void testCreateNew() {
```
mww-sopra-steria, line 94: A newly created ArendeDraft will receive an id from the Repository. Maybe include that in the test.

mww-sopra-steria (changes requested):

web/src/test/java/se/inera/intyg/webcert/web/service/arende/ArendeDraftServiceImplTest.java
mhornfeldt: Is it possible to test this when the return from the repo is being mocked?

mww-sopra-steria: Sure you can. You kind of do this in your new test "shouldReturnSavedArendeDraftFromCreate". The issue there is maybe that it actually just test the scenario when a draft already exists.

You have two behaviours in create(...):
1) If no ArendeDraft exists you create a new one (the new one will receive an id from the db)
2) If a ArendeDraft already exists you should update that one.

Your testCreateNew doesn't really verify the return value. It checks that the repository has been called with the correct arguments. 

It takes a little time to understand the tests because it is not completely clear what they test because they do multiple verification. That is OK, and most pragmatic when it comes to legacy test classes, but I personally find it easier to split into multiple tests with one/fewer asserts per test. 

The test class is quite small so changing it to junit 5 and use nested to do some common mock setup would be quite easy.

You could either do that or you add additional assert(s) to testCreateNew to verify that behaviour #1 also returns the newly created ArendeDraft.

mww-sopra-steria (changes requested):
The code looks good! A comment/answer on the tests to look at.

web/src/test/java/se/inera/intyg/webcert/web/service/arende/ArendeDraftServiceImplTest.java
mhornfeldt: I've split up the tests like you suggested, but I still don't understand how I should try that the database sets and id since I have to mock the return from repository.save()

mww-sopra-steria: Pushed a modification to the test to include the id.

mww-sopra-steria (approved):

PR #910, opened 2023-08-30
INTYGFV-15740: Added contact info for answers

mhornfeldt:


mww-sopra-steria (changes requested):
Unit tests? :-)

mww-sopra-steria (approved):
Approved! 

Check this before merge. Can kontaktinfo be null and the code produces a NPE?

mhornfeldt:
> Approved!
> 
> Check this before merge. Can kontaktinfo be null and the code produces a NPE?

It should be null in the existings tests and they pass correctly

PR #912, opened 2023-09-20
INTYGFV-15866: Implementation to ensure a correct recipient description and to include more certificate types and its recipients.

mikaelaromild1:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/CertificateSignAndSendDescriptionFunctionImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import se.inera.intyg.common.support.facade.model.Certificate;
+import se.inera.intyg.common.ts_bas.support.TsBasEntryPoint;
+import se.inera.intyg.common.ts_diabetes.support.TsDiabetesEntryPoint;
+import se.inera.intyg.infra.security.authorities.AuthoritiesHelper;
+import se.inera.intyg.infra.security.common.model.AuthoritiesConstants;
+import se.inera.intyg.webcert.web.web.controller.facade.dto.ResourceLinkDTO;
+import se.inera.intyg.webcert.web.web.controller.facade.dto.ResourceLinkTypeDTO;
+
+@Component
+public class CertificateSignAndSendDescriptionFunctionImpl implements CertificateSignAndSendDescriptionFunction {
```
mhornfeldt, line 24: I think you can shorten the name by removing "Description". Even though the bug was that the description was wrong, the Function is the whole SignAndSend-functionality and not just the description.

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/CertificateSignAndSendDescriptionFunctionImplTest.java
```diff
@@ -0,0 +1,213 @@
...
+    private final WebCertUserService webCertUserService = mock(WebCertUserService.class);
+
+    @Mock
+    private AuthoritiesHelper authoritiesHelper;
+
+    @InjectMocks
+    private CertificateSignAndSendDescriptionFunctionImpl certificateSignAndSendDescriptionFunction;
+
+    
+    @Test
```
mhornfeldt, line 62: If you want you can probably reuse some code by using functions which take the id returned from EntryPoint.MODULE_ID as a parameter. But it works like this too.

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/CertificateSignAndSendDescriptionFunctionImplTest.java
```diff
@@ -0,0 +1,213 @@
...
+    void shallIncludeSignAndSendCertificateWithDescriptionSkatteverketWhenCertificateTypeIsDb() {
```
mhornfeldt, line 63: Since the test is to test the SignAndSendFunction the test names could be "shallIncludeFunction..." instead if you want to make it shorter.

web/src/test/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/CertificateSignAndSendDescriptionFunctionImplTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+    @Test
+    void shallIncludeSignAndSendCertificateWithDescriptionArbetsformedlingenWhenCertificateTypeIsAf00213() {
+        when(authoritiesHelper.isFeatureActive(AuthoritiesConstants.FEATURE_SIGNERA_SKICKA_DIREKT, Af00213EntryPoint.MODULE_ID))
+            .thenReturn(true);
+        final var certificate = CertificateFacadeTestHelper.createCertificate(Af00213EntryPoint.MODULE_ID, CertificateStatus.UNSIGNED);
+        final var actualAvailableFunctions = certificateSignAndSendDescriptionFunction.get(certificate);
+        assertEquals(actualAvailableFunctions.get().getType(), ResourceLinkTypeDTO.SIGN_CERTIFICATE);
+        assertTrue(
+            actualAvailableFunctions.stream()
```
mhornfeldt, line 98: There's a file called ResourceLinkFacadeTestHelper where you could add this logic to check if the description matches a certain value. In that case you wouldn't have to repeat the logic and we could reuse it in the future.

mhornfeldt (commented):
Nice :) Some comments.

mikaelaromild1:
Thank you for the feedback! I made som changes according to our discussion :)

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/CertificateSignAndSendFunctionImpl.java
```diff
@@ -0,0 +1,119 @@
...
+
+
+    public CertificateSignAndSendFunctionImpl(AuthoritiesHelper authoritiesHelper) {
+        this.authoritiesHelper = authoritiesHelper;
+    }
+
+    @Override
+    public Optional<ResourceLinkDTO> get(Certificate certificate) {
+
+        if (isSignedAndSendDirectly(certificate)) {
```
mhornfeldt, line 44: Should we add a check that if the certificate doesn't have the status UNSIGNED return Optional.empty()? Since only drafts can be signed.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/GetCertificatesAvailableFunctionsImpl.java
```diff
@@ -30,8 +30,6 @@
 import org.springframework.beans.factory.annotation.Autowired;
```
mhornfeldt, line 30: All the tests look good. But you could add tests in the old implementation GetCertificatesAvailableFunction to verify that if your service returns a resource link it's added to the result. If you add this test you would be testing lines 250 and 251. You can look at the tests: shallIncludeSRSMinimizedView and shallExcludeSRSMinimizedView for inspiration.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/CertificateSignAndSendFunctionImpl.java
mikaelaromild1: This check is already made in GetCertificatesAvailableFunctionsImpl so doing this is redundant:

```java

 switch (certificate.getMetadata().getStatus()) {
            case UNSIGNED:
                availableFunctions.addAll(
                    getAvailableFunctionsForDraft(certificate)
                );
                break;
            case SIGNED:
                availableFunctions.addAll(
                    getAvailableFunctionsForCertificate(certificate)
                );
                break;
            case LOCKED:
                availableFunctions.addAll(
                    getAvailableFunctionsForLockedDraft(certificate)
                );
                break;
            case REVOKED:
                availableFunctions.addAll(
                    getAvailableFunctionsForRevokedCertificate(certificate)
                );
                break;
            default:
        }
```

mhornfeldt: The check in AvailableFunctions should exist but now that the code is being refactored it's nice when the logic that is extracted works by itself without logic from the class using it. SrsFunction and CopyCertificateFunction are examples of where the status check is included in the function also. But it works like this too.

mhornfeldt (approved):

PR #918, opened 2023-10-05
INTYGFV-15540: Added recipient to metadata

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/util/CertificateRecipientConverterImpl.java
```diff
@@ -0,0 +1,32 @@
...
+    this.certificateReceiverService = certificateReceiverService;
+  }
+
+  @Override
+  public CertificateRecipient get(String type, String certificateId, LocalDateTime sent) {
+    final var recipients = certificateReceiverService.listPossibleReceiversWithApprovedInfo(type, certificateId);
+
+    return recipients
+        .stream()
+        .findFirst()
```
mww-sopra-steria, line 23: listPossibleReceiversWithApprovedInfo can (in test environments) return multiple recipients. Even if the "huvudmottagare" probably/might be the first element, I would suggest to filter out any receiver that isn't "isLocked" ("huvudmottagare") before findFirst().

web/src/main/java/se/inera/intyg/webcert/web/service/facade/util/UtkastToCertificateConverterImpl.java
```diff
@@ -124,6 +128,13 @@ private Certificate convertToCertificate(Utkast certificate) {
                 certificateToReturn.getMetadata().getTypeVersion())
         );
 
+        certificateToReturn.getMetadata().setRecipient(
+            certificateRecipientConverter.get(
```
mww-sopra-steria, line 132: Use this recipient to replace the code that used for "sentTo" (further up in this class) to find the recipient name. By using the new CertificateRecipientConverter where we fetches this from IT we can get rid of the RecipientConverter (with the hardcoded recipients).

mww-sopra-steria (changes requested):
Take a look at IntygToCertificateConverterImpl as well and add the CertificateRecipient to the metadata.

Also I commented about changing the "sentTo" code to get rid of the hardcoding of recipients in Webcert.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/util/IntygToCertificateConverterImpl.java
```diff
@@ -90,15 +99,18 @@ private Certificate convertToCertificate(IntygContentHolder certificate) {
...
         certificateToReturn.getMetadata().setForwarded(false);
         certificateToReturn.getMetadata().setTestCertificate(certificate.isTestIntyg());
-        certificateToReturn.getMetadata().setSent(
-            certificate.getStatuses().stream().anyMatch(status -> status.getType() == CertificateState.SENT)
+
+        certificateToReturn.getMetadata().setRecipient(
+            certificateRecipientConverter.get(
+                certificate.getUtlatande().getTyp(),
+                certificate.getUtlatande().getId(),
+                sentStatus.map(Status::getTimestamp).orElse(null))
```
mww-sopra-steria, line 107: I like the way you did this!

mww-sopra-steria (approved):

PR #919, opened 2023-10-11
INTYGFV-15540: Added missing licence header, fixed failing tests and …

mikeyOne:
…added a few mor tests


mww-sopra-steria (approved):
Nice with tests for the filter logic! 

Consider renaming the @Nested class from LockStatus -> FilterMainRecipientOnLockStatus (or something similar). This could help out to understand why we filter on LockStatus.

Approved!

mikeyOne:
Ok, will rename class merge

PR #928, opened 2023-11-03
INTYGFV-16034: Set recipient to null for certificates that should not…

mhornfeldt:
… be sent


mww-sopra-steria:
I think it would be good to discuss if we should have this solution in Webcert. We like to control what you can do with a certificate with the use of Resource-links, not to include/exclude information inside the metadata.

PR #929, opened 2023-11-06
INTYGFV-16062: Implementation of GetAvailableFunctionsForCertificateService used by CertificateInteralApiController to get available functions for Mina Intyg.

ChrilleSigerhed:


mww-sopra-steria (changes requested):
We can discuss this on a call as well, but I was thinking that the "smittbärar"-info also should be part of the Print-available function implementation. No need to have it separate as it belongs to printing. Also, shouldn't we just return either PRINT_CERTIFICATE or CUSTOMIZE_PRINT_CERTIFICATE? And sometimes add "Smittbärar"-info together with PRINT_CERTIFICATE when needed.

mww-sopra-steria (approved):
Well done!

PR #931, opened 2023-11-09
INTYGFV-16034: Added available function for send

mhornfeldt:


mww-sopra-steria (approved):
Nice implementation!

You don't really have to add new tests to GetAvailableFunctionsForCertificateServiceTest.java. It is a complete generic class that loads the AvailableFunction-implementations and calls them.

PR #977, opened 2024-02-02
INTYGFV-16340: Refactor certificateServiceProfile management. Added HSA-integration to fetch unit information. Added integration and support to use testability with new CertificateType FK7211.

ChrilleSigerhed:


ChrilleSigerhed:
image

web/src/main/java/se/inera/intyg/webcert/web/web/controller/testability/facade/util/CreateCertificateTestabilityUtil.java
```diff
@@ -235,11 +260,12 @@ private IntygUser getUser(String personId) {
     }
 
     private Patient getPatient(String patientId, String type, String typeVersion) {
```
ChrilleSigerhed, line 262: We get this error from this code if we try to use it ```java.lang.IllegalArgumentException: Unknown intygsTyp: fk7211```

Could we skip this step and instead just:

```java
final var patient = new Patient();
final var personnummer = Personnummer.createPersonnummer(patientId).orElseThrow();  
final var personFromPUService = patientDetailsResolver.getPersonFromPUService(
 personnummer
);
```

web/src/main/java/se/inera/intyg/webcert/web/csintegration/unit/CertificateServiceHsaUnitConverter.java
```diff
@@ -0,0 +1,64 @@
...
+    public CertificateServiceUnitDTO convert(Unit unit) {
+        final var convertedUnit = new CertificateServiceUnitDTO();
+        convertedUnit.setId(unit.getUnitHsaId());
+        convertedUnit.setName(unit.getUnitName());
+        convertedUnit.setAddress(unit.getPostalAddress().isEmpty() ? null : unit.getPostalAddress().get(0));
+        convertedUnit.setZipCode(unit.getPostalCode());
+        convertedUnit.setCity(shouldIncludeCity(unit) ? unit.getPostalAddress().get(unit.getPostalAddress().size() - 1) : null);
+        convertedUnit.setPhoneNumber(unit.getTelephoneNumber().isEmpty() ? null : unit.getTelephoneNumber().get(0));
+        convertedUnit.setEmail(unit.getMail());
+        convertedUnit.setInactive(isActive(unit.getUnitStartDate(), unit.getUnitEndDate()));
```
mww-sopra-steria, line 38: Inactive for the Unit should only be set to true if there is a user session with "integration"-parameter 'inactiveUnit' is set.(https://inera.atlassian.net/wiki/spaces/EIT/pages/6455966/Integration+med+Webcert#4.3.2-Enhet)

And it is only for the unit the user launches to.  You can for now always set it to false and it can be dealt with when its time too implement "DJUPINTEGRATION".

web/src/main/java/se/inera/intyg/webcert/web/csintegration/util/CertificateServiceProfileActive.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package se.inera.intyg.webcert.web.csintegration.util;
+
+import org.springframework.context.annotation.Profile;
+import org.springframework.stereotype.Component;
+
+@Component
+@Profile("certificate-service-active")
+public class CertificateServiceProfileActive implements CertificateServiceProfile {
```
mww-sopra-steria, line 27: Good idea to extract this to a separate component. Easier to reuse and also test! First thought was that it could just be one implementation and it checks for the profile through Environment, but at second thought I kind of like what you done! :-)

web/src/main/java/se/inera/intyg/webcert/web/web/controller/testability/facade/util/CreateCertificateTestabilityUtil.java
```diff
@@ -96,6 +108,19 @@ public String createNewCertificate(@NotNull CreateCertificateRequestDTO createCe
             createCertificateRequest.getCertificateTypeVersion()
         );
 
+        if (certificateServiceProfile.active()) {
```
mww-sopra-steria, line 111: If you can combine the active() and the certificateTypeExits() in one private method, you can "flatten" the code depth.

Also after receiving the modelIdDTO you should also check the version returned if it corresponds to the version to create.

Another thing is to make sure you create the version asked for and not the one returned by certificateTypeExists, as it will always return the latest active version. 

If you want you can deal with these things in this PR, or we manage them in the next "iteration" of the testability api, which will be next sprint when we add content with the certificate.

When I write this I think best would be to take it in the next iteration where we can add a testability endpoint in CS to check a specific version.

web/src/main/java/se/inera/intyg/webcert/web/web/controller/testability/facade/util/CreateCertificateTestabilityUtil.java
mww-sopra-steria: There is some special logic for some legacy types that could be affected if we change this for all. Best would be to find a way to use what you suggest for any certificate types in CS, but the old way for the old types. 

You hardcode a solution for  now and we refactor the solution during next sprint when remove any references to FK7211 in Webcert. But as we have it for now in SupportedCertificateTypesUtil it is fine.

mww-sopra-steria (approved):
Nice work!

A few comments to consider, but not critical for this PR as we will continue to add more things to testability in the next sprint. Have a look and if you agree or have solutions feel free to go forward.

One thing you should do is to move all classes related to CS to 'se.inera.intyg.webcert.web.csintegration', even if this have to do with testability. Good to keep all the logic together relating to CS.

PR #978, opened 2024-02-04
INTYGFV-16344: Replace google autovalue with lombok. Also cleaned up some of the Sonarlint violations.

mww-sopra-steria:


ChrilleSigerhed (approved):
Awsome!!

PR #983, opened 2024-02-07
INTYGFV-16359: Get supported types from certificate-service by request.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great! Nice to see completely generic code in webcert for handling types from CS.

PR #986, opened 2024-02-14
INTYGFV-16375: Implemented possibility to delete certificate stored in certificate service

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/CSIntegrationRequestFactory.java
```diff
@@ -88,6 +89,15 @@ public SaveCertificateRequestDTO saveRequest(Certificate certificate) {
             .build();
     }
 
+    public DeleteCertificateRequestDTO deleteCertificateRequest() {
+        return DeleteCertificateRequestDTO.builder()
```
ChrilleSigerhed, line 93: Please add patient before merging.

mhornfeldt: Can't add patient without getting it communicated from CS, breaking api against frontend, or making extra call against getCertificate in CS

PR #988, opened 2024-02-16
INTYGFV-16387: Merged certificates from certificate service from unsi…

mhornfeldt:
…gned draft list


mikaelaromild1 (approved):

mhornfeldt:
To do: update with filtering when discussed by the team how it will be performed

PR #994, opened 2024-03-05
INTYGFV-16415: Change so that react is default client for user with login method djupintegrerat.

ChrilleSigerhed:


mww-sopra-steria (changes requested):
Nice! Maybe remove the feature-toggling for the react-pilot also as it won't be needed anymore. Maybe its only the ReactClientUtil, but good to double-check.

mww-sopra-steria (approved):
Nice! Approved!

PR #997, opened 2024-03-07
INTYGFV-16443: Added option to hide diagnose for AG114 when printing certificate

mikaelaromild1:


mww-sopra-steria (approved):
Looks good!

PR #1003, opened 2024-03-13
INTYGFV-16435: Access evaluation parameter values for question should…

mhornfeldt:
… be retrieved from certificate


mww-sopra-steria (approved):
Looks good!

PR #1004, opened 2024-03-14
INTYGFV-16426: Implemented support for SITHS signing option.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great! And it's a testament that the "fakesigning" implementation was well structured! 2 down, 1 to go! :-)

PR #1006, opened 2024-03-15
INTYGFV-16427: Implemented BankId Signing option for certificates stored in certificate service.

ChrilleSigerhed:


mww-sopra-steria (approved):
Works great!  

Took the freedom to add some logic to handle private doctors correctly when integrating with CS.

PR #1007, opened 2024-03-18
INTYGFV-16427: Added nullcheck for hosperson fullname.

ChrilleSigerhed:


mhornfeldt (approved):

mhornfeldt:
Test can be added but good without also :)

PR #1008, opened 2024-03-19
INTYGFV-16454: Added functionality to print certificate from cs service

mhornfeldt:


ChrilleSigerhed (approved):
I think this looks great, well done!

Please add the service/component annotation so that the beans can be created correctly!

web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/CSIntegrationService.java
```diff
@@ -285,4 +289,16 @@ public Certificate signCertificateWithoutSignature(SignCertificateWithoutSignatu
...
         return response.getCertificate();
     }
+
+    public IntygPdf printCertificate(String certificateId, PrintCertificateRequestDTO request) {
+        final var url = baseUrl + CERTIFICATE_ENDPOINT_URL + "/" + certificateId + "/pdf";
+
+        final var response = restTemplate.postForObject(url, request, PrintCertificateResponseDTO.class);
+
+        if (response == null) {
+            return null;
```
mww-sopra-steria, line 299: Lets throw exception instead of returning null. Would be great if you could check and add that to any other methods not throwing.

mww-sopra-steria (approved):
Looks good! Please add throws instead of returning null before merging.

PR #1009, opened 2024-03-20
INTYGFV-16451: Refactored user and unit helper to use WebCertUser ins…

mhornfeldt:
…tead of user from facade


mhornfeldt:
Need to check if handling of private doctors is okay

web/src/main/java/se/inera/intyg/webcert/web/csintegration/unit/CertificateServiceUnitHelper.java
```diff
@@ -20,43 +20,55 @@
...
     private final CertificateServiceVardenhetConverter certificateServiceVardenhetConverter;
-    private final HsaOrganizationsService hsaOrganizationsService;
 
-    private static final String PRIVATE_DOCTOR = "Privatläkare";
-
-    public CertificateServiceUnitHelper(UserService userService, CertificateServiceUnitConverter certificateServiceUnitConverter,
-        CertificateServiceVardenhetConverter certificateServiceVardenhetConverter, HsaOrganizationsService hsaOrganizationsService) {
-        this.userService = userService;
-        this.certificateServiceUnitConverter = certificateServiceUnitConverter;
+    public CertificateServiceUnitHelper(WebCertUserService webCertUserService,
```
mww-sopra-steria, line 34: Feel free to use annotation for requiredargsconstuctor in Lombok.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/user/CertificateServiceUserHelper.java
```diff
@@ -84,6 +81,15 @@ private boolean hasSubscription(WebCertUser webCertUser) {
...
 
+    private CertificateServiceUserRole getRole(WebCertUser webCertUser) {
+        final var roles = webCertUser.getRoles();
+        if (roles == null || roles.values().isEmpty()) {
+            throw new IllegalStateException("User has no roles");
+        }
+
+        return convertRole(roles.values().stream().findFirst().orElseThrow().getDesc());
+    }
+
```
mww-sopra-steria, line 92: Now that we use WebCertUser we can also simplify the mapping to role. Consider something like below:

Soon the role of nurse will also exist in AuthoritiesConstants (with this PR: https://github.com/sklintyg/infra/pull/248)

Also it is better to fail fast if we don't recognise the role, so we don't get problems "further in" (read CS).

`
private CertificateServiceUserRole convertRole(String role) {
        switch (role) {
            case AuthoritiesConstants.ROLE_LAKARE:
                return CertificateServiceUserRole.DOCTOR;
            case AuthoritiesConstants.ROLE_PRIVATLAKARE:
                return CertificateServiceUserRole.PRIVATE_DOCTOR;
....
....
....
            default:
                throw new IllegalArgumentException("Dont recognise the role: " + role);
        }
    }
`

mww-sopra-steria: Sorry, the formatting wasn't good when adding code in comment... :-(

mww-sopra-steria (changes requested):
Looks really good! Would be great to simplify the role-conversion as well now that we are using the WebCertUser directly.

PR #1010, opened 2024-03-21
INTYGFV-16460: Added aggregation for send

mhornfeldt:


ChrilleSigerhed (approved):
Nice and clean, well done!

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/SendCertificateFromCertificateService.java
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        monitoringLogService.logIntygSent(
+            certificate.getMetadata().getId(),
+            certificate.getMetadata().getType(),
+            certificate.getMetadata().getRecipient().getId()
+        );
+        pdlLogService.logSent(certificate);
+
+        return certificate.getMetadata().getId();
```
mww-sopra-steria, line 65: The Webcert-implementation returns OK if successfully sent and returns FAILED is not. Double check the implementation by comparing to the Webcert-facade implementation.

mww-sopra-steria (changes requested):
See the mapping of Previous and Signed List. When Sent, it should set the status to "SENT".

mww-sopra-steria (changes requested):
Nice! Works well not in Webcert, but there is another tweak that is necessary.

In CertificateListItemConverterImpl on row 80-82 there needs to be special handling for converting ListIntygEntry to correct status based on the ListType.CERTIFICATES. It should consider the status (matches "SENT") instead of being hardcoded to false (was not introduced in this pr).

mww-sopra-steria (approved):
Nice! Approved!

PR #1014, opened 2024-03-28
INTYGFV-16434: Removed logic related to CertificateEvent since its handled in certificate-service from now on.

ChrilleSigerhed:


mww-sopra-steria (approved):
Nice!

PR #1019, opened 2024-04-07
INTYGFV-16421: Removed environment check for extraction of origin fro…

mikeyOne:
…m request when fake login.


mww-sopra-steria (approved):
Nice!

PR #1026, opened 2024-04-18
INTYGFV-16528: Implementation of CreateDraftCertificateAggregation logic.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/CSIntegrationService.java
```diff
@@ -187,6 +199,33 @@ public Certificate createCertificate(CreateCertificateRequestDTO request) {
...
+    public CreateDraftCertificateResponseType createDraftCertificate(CreateCertificateRequestDTO request, IntygUser user) {
+        final var url = baseUrl + CERTIFICATE_ENDPOINT_URL;
+        try {
+            final var response = restTemplate.postForObject(url, request, CertificateServiceCreateCertificateResponseDTO.class);
+
+            if (response == null) {
+                throw new IllegalStateException(NULL_RESPONSE_EXCEPTION);
+            }
+
+            pdlLogService.logCreatedWithIntygUser(response.getCertificate(), user);
```
mhornfeldt, line 211: Why do we perform the pdl and monitorlog here instead of following the pattern and performing it in the CS-service?

web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/CSIntegrationService.java
```diff
@@ -187,6 +199,33 @@ public Certificate createCertificate(CreateCertificateRequestDTO request) {
         return response.getCertificate();
     }
 
+    public CreateDraftCertificateResponseType createDraftCertificate(CreateCertificateRequestDTO request, IntygUser user) {
```
mhornfeldt, line 202: Maybe I remember wrong but I remember in the design meeting we said we could use the same create endpoint we already have

web/src/main/java/se/inera/intyg/webcert/web/csintegration/aggregate/CreateDraftCertificateAggregator.java
```diff
@@ -0,0 +1,52 @@
...
+import se.inera.intyg.webcert.web.csintegration.certificate.CreateDraftCertificateFromCS;
+import se.inera.intyg.webcert.web.csintegration.util.CertificateServiceProfile;
+import se.inera.intyg.webcert.web.integration.interactions.createdraftcertificate.v3.CreateDraftCertificateFromWC;
+import se.riv.clinicalprocess.healthcond.certificate.createdraftcertificateresponder.v3.CreateDraftCertificateResponseType;
+import se.riv.clinicalprocess.healthcond.certificate.createdraftcertificateresponder.v3.Intyg;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class CreateDraftCertificateAggregator {
```
mhornfeldt, line 35: Do we not want to/can follow the pattern where the aggregator and the services implement the same interface?

web/src/main/java/se/inera/intyg/webcert/web/csintegration/unit/CertificateServiceIntegrationUnitHelper.java
```diff
@@ -0,0 +1,48 @@
...
+import se.inera.intyg.infra.integration.hsatk.model.legacy.AbstractVardenhet;
+import se.inera.intyg.infra.security.common.model.IntygUser;
+
+@Component
+@RequiredArgsConstructor
+public class CertificateServiceIntegrationUnitHelper {
+
+    private final CertificateServiceVardenhetConverter certificateServiceVardenhetConverter;
+
+    public CertificateServiceUnitDTO getCareProvider(IntygUser user) {
```
mhornfeldt, line 33: This file is a copy of the UnitHelper that uses WebcertUser which implement IntygUser. I think it would be better to reuse the code so we don't risk that the to helpers diverge

mhornfeldt: Same for UserHelper

mhornfeldt (commented):
Nice!! Left some comments to discuss next week

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/CreateDraftCertificateFromCS.java
```diff
@@ -0,0 +1,89 @@
...
+                String.format("Cannot create Personnummer object with invalid personId '%s'", personIdExtension)));
+
+        final var sekretessStatus = patientDetailsResolver.getSekretessStatus(personId);
+
+        if (sekretessStatus == SekretessStatus.UNDEFINED) {
+            return CreateDraftCertificateResponseFactory.createErrorResponse(
+                "Information om skyddade personuppgifter kunde inte hämtas. Intyget kan inte utfärdas"
+                    + " för patient när uppgift om skyddade personuppgifter ej är tillgänglig.", ErrorIdType.APPLICATION_ERROR);
+        }
+        return null;
```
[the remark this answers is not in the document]
ChrilleSigerhed: Good idea, using an Optional looks better!

web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/CSIntegrationService.java
ChrilleSigerhed: Valid point, I will correct this error.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/unit/CertificateServiceIntegrationUnitHelper.java
ChrilleSigerhed: In then first iteration I reused the code with this function:
```java
    private IntygUser getUser(Optional<IntygUser> optionalUser) {
        return optionalUser.orElseGet(webCertUserService::getUser);
    }
```

But from my understanding the request will diff more after we introduce more parameters related to the deep integrated user, therefor it was my understanding that it might be a good idea to separate them.

I will look into this and update if we can reuse more code!

web/src/main/java/se/inera/intyg/webcert/web/csintegration/aggregate/CreateDraftCertificateAggregator.java
ChrilleSigerhed: I think its possible to implement the interface ```CreateDraftCertificateResponderInterface```, but I decided to not do it in the first iteration because I thought it felt "forced". 

Mainly because we decided to leave alot of the common validation in ```CreateDraftCertificateResponderImpl```. And if we decided to implement the interface we would end up with four implementation, wich felt wrong to me. 

But I might be wrong and we should indeed implement the interface.

mhornfeldt: I was more thinking of creating an interface like "CreateDraftCertificiate" to keep the pattern we've used in previous aggregator implementations, and make sure that the cs and wc implementation doesn't start to diverge. But maybe it's not the way to go

ChrilleSigerhed: Ah I see! I did not think of that

I like the suggestion and will add it to the PR!

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/CreateCertificateFromCertificateService.java
```diff
@@ -62,6 +64,14 @@ private String createCertificate(String patientId, CertificateModelIdDTO modelId
         );
 
         pdlLogService.logCreated(response);
+        monitoringLogService.logUtkastCreated(
+            response.getMetadata().getId(),
+            response.getMetadata().getType(),
+            response.getMetadata().getUnit().getUnitId(),
+            response.getMetadata().getIssuedBy().getPersonId(),
+            0
```
mww-sopra-steria, line 72: Suggest to create a constant for the "0". Call it something like NO_PREFILL_ELEMENTS.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/CreateDraftCertificateFromCS.java
```diff
@@ -0,0 +1,115 @@
...
+                csIntegrationRequestFactory.createDraftCertificateRequest(modelId.get(), certificate, user)
+            );
+
+            pdlLogService.logCreatedWithIntygUser(createdCertificate, user);
+            monitoringLogService.logUtkastCreated(
+                createdCertificate.getMetadata().getId(),
+                createdCertificate.getMetadata().getType(),
+                createdCertificate.getMetadata().getUnit().getUnitId(),
+                createdCertificate.getMetadata().getIssuedBy().getPersonId(),
+                0
```
mww-sopra-steria, line 83: Suggest to create a constant for the "0". Call it something like NO_PREFILL_ELEMENTS.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/CreateDraftCertificateFromCS.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private static IntegreradEnhetEntry getIntegreradEnhetEntry(IntygUser user) {
+        return new IntegreradEnhetEntry(user.getValdVardenhet().getId(),
+            user.getValdVardenhet().getNamn(), user.getValdVardgivare().getId(), user.getValdVardgivare().getNamn());
+    }
+
+    private Optional<CreateDraftCertificateResponseType> validatePUIntegration(Intyg certificate) {
+        final var personIdExtension = certificate.getPatient().getPersonId().getExtension();
+        final var personId = Personnummer.createPersonnummer(personIdExtension)
+            .orElseThrow(() -> new IllegalArgumentException(
```
mww-sopra-steria, line 100: Webcert throws WebCertServiceException if it is not possible. Use the same type of exception and type.

`        final var personnummer = Personnummer.createPersonnummer(
                certificate.getPatient().getPersonId().getExtension())
            .orElseThrow(() -> new WebCertServiceException(WebCertServiceErrorCodeEnum.PU_PROBLEM,
                "Failed to create valid personnummer for createDraft request"));`

mww-sopra-steria (approved):
Looks good and the refactoring of the old code seems to be done with care. 

A few minor comments to consider before merge.

Good discussion and valid point from @mhornfeldt concerning the "request-factory". I would say, go forward with this PR and assess this after the "djupintegration"-parameters have been handled. It is good if we don't repeat ourself, but at the same time trying to keep things simple.

PR #1027, opened 2024-04-18
INTYGFV-16478: Added login methods ccp19 and ccp28 for mobile bankid.…

mikeyOne:
… Updated service name for subscription service call.


mww-sopra-steria (approved):
Looks good!

PR #1035, opened 2024-04-26
INTYGFV-16545: Added IntegrationService aggregator to handle deep int…

ChrilleSigerhed:
…egration for certificate created in CertificateService.

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/AlternateSsnEvaluator.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/AlternateSsnEvaluator.java
@@ -33,15 +33,15 @@ public class AlternateSsnEvaluator {
             return false;
         }
 
-        if (!alternateSsnProvidedAndDoesNotMatchPatientIdOnCertificate(certificate, user)) {
+        if (alternateSsnProvidedAndDoesMatchPatientIdOnCertificate(certificate, user)) {
             return false;
         }
 
         return certificate.getLinks().stream().anyMatch(link -> link.getType().equals(ResourceLinkTypeEnum.EDIT_CERTIFICATE));
     }
 
-    private static boolean alternateSsnProvidedAndDoesNotMatchPatientIdOnCertificate(Certificate certificate, WebCertUser user) {
+    private static boolean alternateSsnProvidedAndDoesMatchPatientIdOnCertificate(Certificate certificate, WebCertUser user) {
         return user.getParameters() != null && user.getParameters().getAlternateSsn() != null && !user.getParameters().getAlternateSsn()
-            .isBlank() && !user.getParameters().getAlternateSsn().equals(certificate.getMetadata().getPatient().getPersonId().getId());
+            .isBlank() && user.getParameters().getAlternateSsn().equals(certificate.getMetadata().getPatient().getPersonId().getId());
     }
 }
--- a/web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/AlternateSsnEvaluator.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/AlternateSsnEvaluator.java
@@ -44,10 +44,6 @@ public class AlternateSsnEvaluator {
             return false;
         }
 
-        if (Personnummer.createPersonnummer(user.getParameters().getAlternateSsn()).orElse(null) == null) {
-            return false;
-        }
-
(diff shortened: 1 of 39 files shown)


mww-sopra-steria (changes requested):
Nice work! This was not an easy task but you managed really well!

I have a few comments that need to be looked at. Its a bit tricky to interpret the current logic and lets discuss if something doesn't make sense.

Also I have a bit of concern relating to the prepareRedirectToIntyg(...) on line 355 in intygIntegrationController. I'm not sure it will work correctly as it is. When we don't have a unit selected. Lets talk about it.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/AlternateSsnEvaluator.java
```diff
@@ -0,0 +1,66 @@
...
+        if (alternateSsnProvided(user)) {
+            return false;
+        }
+
+        if (removeDash(user.getParameters().getAlternateSsn()).equals(
+            removeDash(certificate.getMetadata().getPatient().getPersonId().getId()))) {
+            return false;
+        }
+
+        if (Personnummer.createPersonnummer(user.getParameters().getAlternateSsn()).orElse(null) == null) {
```
mww-sopra-steria, line 47: Works fine, but you could just do isEmpty() instead of the orElse(null) == null which is not as readable.

You could also move this to "alternateSsnProvided" which will make sense.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/CertificateDetailsUpdateService.java
```diff
@@ -0,0 +1,75 @@
...
+                )
+            );
+        }
+
+        if (shouldUpdatePatientDetails) {
+            monitoringLogService.logUtkastPatientDetailsUpdated(
+                certificate.getMetadata().getId(),
+                certificate.getMetadata().getType()
+            );
+            publishCertificateStatusUpdateService.publish(certificate, HandelsekodEnum.ANDRAT);
```
mww-sopra-steria, line 65: We should publish the certificate that has been saved. Otherwise it won't contain the new patient details and the externalReference.

web/src/main/java/se/inera/intyg/webcert/web/web/controller/integration/IntygIntegrationController.java
```diff
@@ -434,10 +413,10 @@ private Response buildErrorResponse(UriInfo uriInfo, String errorReason) {
         return Response.temporaryRedirect(location).build();
     }
 
-    private Response buildChooseUnitResponse(UriInfo uriInfo, PrepareRedirectToIntyg prepareRedirectToIntyg) {
+    private Response buildChooseUnitResponse(UriInfo uriInfo, String certificateId) {
```
mww-sopra-steria, line 416: Shouldn't we have the intygtype as well? And just ignore the intygtypeversion?

web/src/main/java/se/inera/intyg/webcert/web/web/controller/integration/dto/PrepareRedirectToIntyg.java
```diff
@@ -19,43 +19,19 @@
...
-    public boolean isUtkast() {
-        return utkast;
-    }
-
-    public void setUtkast(boolean utkast) {
-        this.utkast = utkast;
-    }
 
     @Override
     public String toString() {
```
mww-sopra-steria, line 37: Do we need to keep the toString?

mww-sopra-steria (changes requested):
Most things are OK, but please check the comments.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/CertificateDetailsUpdateService.java
ChrilleSigerhed: Updated so the logging also is done on the savedCertificate.

web/src/main/java/se/inera/intyg/webcert/web/web/controller/integration/dto/PrepareRedirectToIntyg.java
ChrilleSigerhed: Removed it

mww-sopra-steria (approved):
Looks good!

PR #1036, opened 2024-04-29
INTYGFV-16292: Fixed issue when creating draft from createDraftCertificate when no user is logged in.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #1037, opened 2024-05-02
INTYGFV-16556: Added aggregation for replace certificate

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/ReplaceCertificateFromCertificateService.java
```diff
@@ -0,0 +1,69 @@
...
+        );
+
+        if (replacingCertificate == null) {
+            throw new IllegalStateException("Received null when trying to replace certificate from Certificate Service");
+        }
+
+        log.debug("Replaced certificate '{}' from Certificate Service", certificateId);
+        monitoringLogService.logIntygCopiedReplacement(replacingCertificate.getMetadata().getId(), certificateId);
+        pdlLogService.logCreated(replacingCertificate);
+
```
ChrilleSigerhed, line 66: We need to publish the event aswell!

```java
publishCertificateStatusUpdateService.publish(createdCertificate, HandelsekodEnum.SKAPAT);
```

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/ReplaceCertificateFromCertificateService.java
```diff
@@ -0,0 +1,69 @@
...
+    private final CSIntegrationService csIntegrationService;
+    private final CSIntegrationRequestFactory csIntegrationRequestFactory;
+    private final PDLLogService pdlLogService;
+    private final MonitoringLogService monitoringLogService;
+
+    @Override
+    public String replaceCertificate(String certificateId) {
+        log.debug("Attempting to replace certificate '{}' from Certificate Service", certificateId);
+
+        final var certificateToReplace = csIntegrationService.getCertificate(
```
ChrilleSigerhed, line 45: Get certificate throws if the certificate is not found in Certificate Service. From CS:

```java
 @Override
  public Certificate getById(CertificateId certificateId) {
    if (certificateId == null) {
      throw new IllegalArgumentException("Cannot get certificate if certificateId is null");
    }

    final var certificateEntity = certificateEntityRepository.findByCertificateId(
            certificateId.id())
        .orElseThrow(() ->
            new IllegalArgumentException(
                "CertificateId '%s' not present in repository".formatted(certificateId)
            )
        );

    return certificateEntityMapper.toDomain(certificateEntity);
  }
```

Therefore I think we should first check if the certificate exists, before trying to fetch it.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/aggregate/ReplaceCertificateAggregator.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package se.inera.intyg.webcert.web.csintegration.aggregate;
+
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.stereotype.Service;
+import se.inera.intyg.webcert.web.csintegration.util.CertificateServiceProfile;
+import se.inera.intyg.webcert.web.service.facade.ReplaceCertificateFacadeService;
+
+@Service("replaceCertificateAggregator")
```
ChrilleSigerhed, line 27: Add this qualifier to ```ReplaceCertificateFacadeService``` in ```CertificateController``` aswell.

ChrilleSigerhed (changes requested):
Nice, well done! Left some comments to consider.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/ReplaceCertificateFromCertificateService.java
```diff
@@ -0,0 +1,78 @@
...
+public class ReplaceCertificateFromCertificateService implements ReplaceCertificateFacadeService {
+
+    private final CSIntegrationService csIntegrationService;
+    private final CSIntegrationRequestFactory csIntegrationRequestFactory;
+    private final PDLLogService pdlLogService;
+    private final MonitoringLogService monitoringLogService;
+    private final PublishCertificateStatusUpdateService publishCertificateStatusUpdateService;
+
+    @Override
+    public String replaceCertificate(String certificateId) {
```
mww-sopra-steria, line 44: I looked at the current Webcert-code when user is replacing a certificate. There is logic to register the unit if the user is in origin DJUPINTEGRATION. See CopyUtkastServiceImpl and method checkIntegreradEnhet(...).

We have to handle it here as well.

mww-sopra-steria (changes requested):
I noticed something we haven't considered, during the design meeting, when I was reading Webcert code. See comment.

Would suggest that you go through the "replace"-code in current implementation as well to see so we don't miss anything else. There are so many details in Webcert that are hard to catch without some "code archeology" :-)

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/IntegratedUnitHelper.java
```diff
@@ -0,0 +1,51 @@
...
+import se.inera.intyg.webcert.web.service.user.WebCertUserService;
+
+@Component
+@RequiredArgsConstructor
+public class IntegratedUnitHelper {
+
+    private final WebCertUserService webCertUserService;
+    private final IntegreradeEnheterRegistry integreradeEnheterRegistry;
+
+    public void registerUnit() {
```
ChrilleSigerhed, line 36: Nice helper method to handle storage of integrated units. Two questions

1. If we change the name of the class to IntegratedUnitRegistryHelper, does it make the purpose of the class clearer? If you agree, maybe update the method-name to register.
2. Would be nice if we could use this helper in ```CreateDraftCertificateFromCS``` aswell (on line 78), do you think its worth refactoring so that we can handle ```IntygUser``` aswell?

mww-sopra-steria (approved):
Looks good! And great that you added externalReference to the request as well.

I notice a little difference how the "CS-code" update "IntegratedUnit". Before merging, just doublecheck that we are getting the correct behaviour.

ChrilleSigerhed (approved):
Great job!

mhornfeldt:
> Looks good! And great that you added externalReference to the request as well.
> 
> I notice a little difference how the "CS-code" update "IntegratedUnit". Before merging, just doublecheck that we are getting the correct behaviour.

Which difference do you refer to?

mww-sopra-steria:
See See CopyUtkastServiceImpl and method checkIntegreradEnhet(...) and compare to the CS-code.

PR #1039, opened 2024-05-02
INTYGFV-16548: Added logic to handle external type when fetching certificate from certificate-service.

ChrilleSigerhed:


mikeyOne (approved):
Looks good!

PR #1040, opened 2024-05-03
INTYGFV-16560: Added aggregator for renew

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/RenewCertificateFromCertificateService.java
```diff
@@ -0,0 +1,94 @@
...
+        if (certificateToRenew == null) {
+            throw new IllegalStateException(
+                String.format("Certificate service returned null when getting certificate '%s'", certificateId)
+            );
+        }
+
+        final var integrationParameters = webCertUserService.getUser().getParameters();
+        final var renewalCertificate = csIntegrationService.renewCertificate(
+            certificateId,
+            csIntegrationRequestFactory.renewCertificateRequest(
```
mww-sopra-steria, line 69: Should this be part of the csIntegrationRequestFactory.renewCertificateRequest instead? What I mean is that the "alternateSsn" and "reference" code could be part of the request factory.

The logic needs to be managed for multiple type of requests, today we know that it is renew and replace.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/RenewCertificateFromCertificateService.java
```diff
@@ -0,0 +1,94 @@
...
+                isAlternateSSNDefined(integrationParameters) ? integrationParameters.getAlternateSsn()
+                    : certificateToRenew.getMetadata().getPatient().getPersonId().getId(),
+                integrationParameters != null ? integrationParameters.getReference() : null
+            )
+        );
+
+        if (renewalCertificate == null) {
```
mww-sopra-steria, line 76: I've been thinking about this a few times before... it is really relevant to do null check on the return-value? Shouldn't you be able to trust the renewCertificate to always return a value and if something is wrong always throw an exception?

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/RenewCertificateFromCertificateService.java
```diff
@@ -0,0 +1,94 @@
...
+            throw new IllegalStateException(
+                String.format("Certificate service returned null when getting certificate '%s'", certificateId)
+            );
+        }
+
+        final var integrationParameters = webCertUserService.getUser().getParameters();
+        final var renewalCertificate = csIntegrationService.renewCertificate(
+            certificateId,
+            csIntegrationRequestFactory.renewCertificateRequest(
+                isAlternateSSNDefined(integrationParameters) ? integrationParameters.getAlternateSsn()
```
mww-sopra-steria, line 70: I found some code in the current facade-implementation that manages a situation if the personId is a "reserve id". See line 71 in RenewCertificateFacadeServiceImpl. Good to look at the whole commit that includes changes in multiple classes.

mww-sopra-steria (changes requested):
Looks good! Have a few refactoring suggestion. Also I noticed some special handling of reserve id in current Webcert-code that we should follow up on.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/RenewCertificateFromCertificateService.java
mww-sopra-steria: How about this question? :-)

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/RenewCertificateFromCertificateService.java
mww-sopra-steria: No, I don't think it is necessary. Any reserveIds handled by Webcert due to IntegrationParameters should stay in Webcert.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/RenewCertificateFromCertificateService.java
mhornfeldt: I had two so I removed the first one only first. Now both are gone.

mww-sopra-steria (approved):
Looks good!

PR #1043, opened 2024-05-08
INTYGFV-16386: Updated check for blocked unit to include hsaId:S of c…

mikeyOne:
…are provider and care unit


mww-sopra-steria (approved):
Looks good!

PR #1046, opened 2024-05-15
INTYGFV-16576: Include legitimate professionalRoles in CertificateServiceUserDTO.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/testability/CertificateServiceUserBuilder.java
```diff
@@ -55,6 +55,7 @@ public CertificateServiceUserDTO build(HoSPersonal hoSPersonal) {
             .role(role)
             .blocked(false)
             .allowCopy(true)
+            .healthCareProfessionalLicence(List.of(LAKARE))
```
mww-sopra-steria, line 58: Shouldn't we use the "licenses" from the user-object?

mww-sopra-steria (changes requested):
Nice that it was so easy to add! See comment though. Would be great to have a solution so not all users are treated as "Doctor" when creating certificates from testability.

mww-sopra-steria (approved):

PR #1048, opened 2024-05-16
INTYGFV-16582: Should send certificate after signing if SEND_AFTER_SIGN_CERTIFICATE ResourceLinkTypeDTO is present.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks really good! Well done!

PR #1051, opened 2024-05-17
INTYGFV-16586: Implemented aggregator for ComplementCertificateFacadeService.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/ComplementCertificateFromCertificateService.java
```diff
@@ -0,0 +1,107 @@
...
+            certificateId,
+            csIntegrationRequestFactory.getCertificateRequest()
+        );
+
+        final var certificate = csIntegrationService.complementCertificate(
+            certificateId,
+            csIntegrationRequestFactory.complementCertificateRequest(
+                certificateToComplement.getMetadata().getPatient(),
+                webCertUser.getUser().getParameters(),
+                message)
```
mww-sopra-steria, line 67: I don't think it is necessary to add the message when complementing. It is legacy logic related the other FK-certificates. Suggest that we remove it from complement request to CS as it won't use it.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/ComplementCertificateFromCertificateService.java
```diff
@@ -0,0 +1,107 @@
...
+        if (Boolean.FALSE.equals(csIntegrationService.certificateExists(certificateId))) {
+            log.debug("Certificate '{}' does not exist in certificate service", certificateId);
+            return null;
+        }
+
+        final var certificate = csIntegrationService.answerComplementOnCertificate(
+            certificateId,
+            csIntegrationRequestFactory.answerComplementOnCertificateRequest(message)
+        );
+
```
mww-sopra-steria, line 96: We should send a status update for the certificate that an answer has been sent. See ArendeServiceImpl.answerKomplettering

mww-sopra-steria (changes requested):
Looks good! Take a look at the comments, most important is that we send status updates.

mww-sopra-steria (approved):
Perfect!

PR #1052, opened 2024-05-20
INTYGFV-16587: Implemented aggregator sendMessageToCare.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/message/SendCertificateQuestionUpdateService.java
```diff
@@ -0,0 +1,72 @@
...
+import se.inera.intyg.webcert.persistence.arende.model.ArendeAmne;
+import se.inera.intyg.webcert.web.integration.registry.IntegreradeEnheterRegistry;
+import se.inera.intyg.webcert.web.service.mail.MailNotification;
+import se.inera.intyg.webcert.web.service.mail.MailNotificationService;
+import se.riv.clinicalprocess.healthcond.certificate.sendMessageToCare.v2.SendMessageToCareType;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class SendCertificateQuestionUpdateService {
```
mww-sopra-steria, line 35: Maybe we should call it something more specific? It makes mail notifications for received messages. Maybe something like SendMailNotificationForReceivedMessageService (maybe a bit long... but you get the point :-) )

web/src/main/java/se/inera/intyg/webcert/web/csintegration/message/dto/SentByDTO.java
```diff
@@ -0,0 +1,23 @@
...
+        this.code = code;
+    }
+
+    public static SentByDTO getByCode(String type) {
+        for (SentByDTO sentByDTO : values()) {
+            if (sentByDTO.getCode().equals(type)) {
+                return sentByDTO;
+            }
+        }
+        return null;
```
mww-sopra-steria, line 21: Throw exception if no match is found?

mww-sopra-steria (approved):
Looks great! See two comments to consider before merge.

PR #1056, opened 2024-05-27
DEVENV: Added integration test resources and fix for test null pointe…

mikeyOne:
…r exception.


mww-sopra-steria (approved):
Nice find and fix!

PR #1064, opened 2024-06-04
INTYGFV-16653: PublishCertificateStatusUpdateService should use internal api to fetch XML data. Should not include HSA_ID if called from internal system when building notification.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/dto/InternalCertificateXmlResponseDTO.java
```diff
@@ -0,0 +1,39 @@
...
+
+package se.inera.intyg.webcert.web.csintegration.integration.dto;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import lombok.Builder;
+import lombok.Value;
+import se.inera.intyg.webcert.web.csintegration.integration.dto.InternalCertificateXmlResponseDTO.InteralCertificateXmlResponseDTOBuilder;
+
+@JsonDeserialize(builder = InteralCertificateXmlResponseDTOBuilder.class)
```
mww-sopra-steria, line 28: Spelling mistake Interal -> Internal

mww-sopra-steria (approved):
Looks good! Check spelling of builder in dto before merging.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/integration/dto/InternalCertificateXmlResponseDTO.java
ChrilleSigerhed: Good catch!

PR #1065, opened 2024-06-04
INTYGFV-16653: Added mapping for RENEW_CERTIFICATE in listview.

ChrilleSigerhed:
image

### DIFF
--- a/web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/ResourceLinkListHelperImpl.java
+++ b/web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/ResourceLinkListHelperImpl.java
@@ -173,6 +173,9 @@ public class ResourceLinkListHelperImpl implements ResourceLinkListHelper {
     }
 
     private boolean validateRenew(ActionLink link, String certificateType, CertificateRelations relations, CertificateStatus status) {
+        if (link.getType() == ActionLinkType.FORNYA_INTYG_FRAN_CERTIFICATE_SERVICE) {
+            return true;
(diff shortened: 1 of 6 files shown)


web/src/main/java/se/inera/intyg/webcert/web/service/facade/list/ResourceLinkListHelperImpl.java
```diff
@@ -174,8 +176,12 @@ private ResourceLinkDTO getConvertedResourceLink(
     }
 
     private boolean validateRenew(ActionLink link, String certificateType, CertificateRelations relations, CertificateStatus status) {
-        return link.getType() == ActionLinkType.FORNYA_INTYG
-            && CertificateRenewFunction.validate(certificateType, relations, status, authoritiesHelper);
+        final var actionLinkIsRenewCertificate = link.getType() == ActionLinkType.FORNYA_INTYG;
```
mww-sopra-steria, line 179: Ah, tricky that we have to handle the situation when we have a correct link and when we have to validate it with CertificateRenewFunction.validate(...).

Now we will call cs everytime we need to check activeAndSupportsType. This will be costly as it will be done in a loop and  for each certificate. Could we "rearrange" things to remove the need of calling cs?

mww-sopra-steria (changes requested):
Nicely done!  A little concern about what we are forced to do in ResourceLinkListHelperImpl... Could we see if we can find a different solution that doesn't require calling cs here?

mww-sopra-steria (approved):
Nice work!

PR #1066, opened 2024-06-12
INTYGFV-16667: Implemented Aggregator for send-, save-,delete- & create Question. Refactored so that handle, getQuestions & getComplementQuestions returns object Question instead of QuestionDTO.  

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great! Well done!

PR #1069, opened 2024-06-19
INTYGFV-16669: Added patient to CreateMessageRequestDTO. Added CertificateId to QuestionDTO.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looking good! 👍

PR #1070, opened 2024-06-20
INTYGFV-16668: Implemented aggregator to enable save, delete and send for answer.

ChrilleSigerhed:


mikaelaromild1 (approved):
Looks good! :)

mhornfeldt (commented):
The exists is placed on different levels if you compare delete to the rest. Do we want to place them on the same level?

PR #1074, opened 2024-07-05
INTYGFV-16519: Added url to forwarded drafts and questions in configuration

mikaelaromild1:


devops/dev/config/application-dev.properties
```diff
@@ -54,6 +54,7 @@ webcert.host.url=https://wc.localtest.me
 webcert.host.url.internal=http://localhost:${dev.http.port}
 sakerhetstjanst.saml.entityId=${webcert.host.url}
 certificate.view.host.react.client=wc2.wc.localtest.me
+forward.draft.or.question.url=${webcert.host.url}/webcert/web/user/launch/certificate/
```
ChrilleSigerhed, line 57: Since this URL will not change depending on enviroment (excluding the webcert.host.url that is already handled) I suggest we move this configuration to application.properties.

ChrilleSigerhed (changes requested):
Good job, left one comment to consider.

ChrilleSigerhed (approved):
Nice, well done!

PR #1080, opened 2024-07-18
INTYGFV-16739: Implemented logic to fetch certificates from CertificateService aswell in ListCertificatesForCareWithQAResponderInterface implementation.

ChrilleSigerhed:


mikaelaromild1 (approved):
Approved through live review

PR #1083, opened 2024-08-09
INTYGFV-16750: Fixed errors related to aggregation of signed certific…

mhornfeldt:
…ate list


mww-sopra-steria (approved):
Nice solution!

PR #1085, opened 2024-08-12
INTYGFV-16638: Shall not get intygData for certificates stored in certificate-service.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/service/intyg/IntygServiceImpl.java
```diff
@@ -685,6 +691,10 @@ public List<IntygWithNotificationsResponse> listCertificatesForCareWithQA(IntygW
         final var draftMap = getDraftMap(notificationCertificateIdHash.keySet());
 
         for (var certificateId : notificationCertificateIdHash.keySet()) {
+            if (certificateServiceProfile.active() && Boolean.TRUE.equals(csIntegrationService.certificateExists(certificateId))) {
```
mww-sopra-steria, line 694: Would be great if we can find a way to check if the certificate is from CS without calling one at a time inside the loop. Maybe check it with one call or reorganise the logic so we don't have to call CS here. Not sure what the best solution is at the moment, but we should go to length not to make this call inside the loop.

mww-sopra-steria (changes requested):
Understand the need for the fix! See comment for feedback on the way we fix this.

web/src/main/java/se/inera/intyg/webcert/web/service/intyg/IntygServiceImpl.java
ChrilleSigerhed: I updated the logic so that we instead use the already retrieved list from certificate-service to determine if the certificate is from cs or not.

mww-sopra-steria (approved):
Good choice of solution!

PR #1087, opened 2024-08-13
INTYGFV-16761: Implemented a solution that retrieves statistics from certificate service and adds it to the existing statistics already retrieved from Webcert.

ChrilleSigerhed:


mhornfeldt (approved):
Nice!

mhornfeldt (approved):

PR #1100, opened 2024-08-30
INTYGFV-16796: Implementation of email notification enabled for deep integrated users.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/GetUnitNotificationConfig.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Slf4j
+@Component
+@RequiredArgsConstructor
+public class GetUnitNotificationConfig {
```
mww-sopra-steria, line 36: Would be great if we could initialise it once and then keep in memory. It ain't a lot of data so it won't be an issue, but it will speed things up if it doesn't have to initialise every time it is requested. 

Also we won't have repeated logs that it was loaded or failed to load.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/GetUnitNotificationConfig.java
```diff
@@ -0,0 +1,57 @@
...
+        final var objectMapper = new ObjectMapper();
+        objectMapper.registerModule(new JavaTimeModule());
+        try (final var resourceAsStream = new FileInputStream(unitNotificationConfigPath)) {
+            final var integratedUnitNotificationConfig = objectMapper.readValue(
+                resourceAsStream,
+                new TypeReference<List<RegionNotificationConfig>>() {
+                });
+            log.info("Integrated Unit Notification was loaded with configuration: {}", integratedUnitNotificationConfig);
+            return integratedUnitNotificationConfig;
+        } catch (Exception e) {
```
mww-sopra-steria, line 52: Could we catch the exception if resource is missing and just log the info explaining it is missing.

All other exceptions should probably log an error and full stacktrace, because it is due to parsing errors.

web/src/main/java/se/inera/intyg/webcert/web/csintegration/message/ProcessIncomingMessageService.java
```diff
@@ -65,8 +70,15 @@ public SendMessageToCareResponseType process(SendMessageToCareType sendMessageTo
             isAnswer
         );
 
-        publishCertificateStatusUpdateService.publish(certificate, getEventType(questionType, isAnswer));
-        sendMailNotificationForReceivedMessageService.send(sendMessageToCare, certificate);
+        final var shouldReciveMailNotifications = shouldReciveMailNotifications(certificate);
+
+        if (unitIsIntegrated(certificate) && !shouldReciveMailNotifications) {
```
mww-sopra-steria, line 75: You could probably simplify this a little bit. For example... You check first if there should be mail-notifications and if not do publish. Would be easier to understand the code and only one thing will be executed. Ex:

`
if (!unitIsIntegrated(certificate) || shouldReciveMailNotifications) {
    sendMailNotificationForReceivedMessageService.send(sendMessageToCare, certificate);
} else {
   publishCertificateStatusUpdateService.publish(certificate, getEventType(questionType, isAnswer));
}
`

web/src/main/java/se/inera/intyg/webcert/web/service/arende/ArendeServiceImpl.java
```diff
@@ -216,6 +220,9 @@ public Arende processIncomingMessage(Arende arende) {
             monitoringLog.logArendeReceived(certificateId, utkast.getIntygsTyp(), utkast.getEnhetsId(), arende.getAmne(),
                 arende.getKomplettering().stream().map(MedicinsktArende::getFrageId).collect(Collectors.toList()),
                 arende.getSvarPaId() != null);
+            Arende saved = arendeRepository.save(arende);
+            sendNotificationAndCreateEventForIncomingMessage(saved, utkast.getVardgivarId(), utkast.getSignatur().getSigneringsDatum());
+            return saved;
```
mww-sopra-steria, line 225: If you like you can remove the "else" as the if contains a return. This will simplify the code a little bit.

web/src/test/java/se/inera/intyg/webcert/web/csintegration/certificate/GetUnitNotificationConfigTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+import java.nio.file.Paths;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.test.util.ReflectionTestUtils;
+
+@ExtendWith(MockitoExtension.class)
+class GetUnitNotificationConfigTest {
```
mww-sopra-steria, line 34: Would be great to include test to assert that the data has been correctly parsed. Thinking about the actual values of the json.

mww-sopra-steria (changes requested):
Really nice and clean! I have a few comments to look at.

mww-sopra-steria (approved):
Looks great! Well done!

PR #1101, opened 2024-09-01
INTYGFV-16799: Add (back!) legacy launch controllers. Removed unnecessary legacy code.

mww-sopra-steria:


mikeyOne (approved):
Looks good

PR #1108, opened 2024-09-04
K1J-95: Restructure some errorlogs when sending email notifications for incoming questions/answers

mww-sopra-steria:
Included inactivation of lock draft job as we have done it in all our test environments.


ChrilleSigerhed (approved):
Very nice, a great improvement to our logs!

PR #1109, opened 2024-09-04
K1J-148: Added origin to user object

mikeyOne:


mww-sopra-steria (approved):
I don't see any better solution myself than adding origin right now. Looks good!

PR #1110, opened 2024-09-10
K1J-146: Updated logic in IntegratedUnitNotificationEvaluator so that it complies with business logic.

ChrilleSigerhed:


mww-sopra-steria (approved):
Nice!

PR #1113, opened 2024-09-10
K1J-183: Updated content of additional info when PDL-logging logSendIntygToRecipient.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks good!

PR #1121, opened 2024-09-18
K1J-232: Updated so additional info only is added to logs if user has authentication context.

ChrilleSigerhed:


mww-sopra-steria (approved):
Nice!

PR #1122, opened 2024-09-19
K1J-165-1: All certificates from CS should be decorated with information from WC

mikaelaromild1:


mhornfeldt (approved):

mww-sopra-steria (approved):
Looks good!

PR #1126, opened 2024-10-10
K1J-371-1: Updated converter for message requests to set subject from rubrik instead of amne

mikaelaromild1:


mww-sopra-steria (approved):
Nice!

PR #1127, opened 2024-10-10
K1J-399: CertificateStatusUpdateForCare should set number of questions for certificates from CS

mikaelaromild1:


ChrilleSigerhed (approved):
Nice!

PR #1134, opened 2024-10-16
K1J-421: Handle certificates from CS in InternalNotificationMessageListener.

ChrilleSigerhed:


mww-sopra-steria (approved):
Looks great! Nice and sweet solution!

PR #1139, opened 2024-10-29
K1J-404: Handle if user has agreement to manage certificates or not. Only users with agreements will be able to work with certificates with the exceptions for reading, deleting, revoking and printing.

mww-sopra-steria:


ChrilleSigerhed (approved):
Well done, nice that you introduced agreement!

PR #1140, opened 2024-10-29
K1J-435: Added has for patient id to not be logged as is

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/logging/HashPatientIdHelper.java
```diff
@@ -0,0 +1,65 @@
...
+import java.security.NoSuchAlgorithmException;
+import java.util.regex.Pattern;
+
+public class HashPatientIdHelper {
+
+    private HashPatientIdHelper() {
+
+    }
+
+    public static String fromUrl(String url) {
```
ChrilleSigerhed, line 33: We have a HashUtility-class defined in schemas that we used for other scenarios when we needed to hash personalId. Should we extend that class with the function ```fromUrl``` instead of creating a new one?

ChrilleSigerhed (changes requested):
Nice, left a comment to consider, let me know what you think!

web/src/main/java/se/inera/intyg/webcert/web/logging/HashPatientIdHelper.java
mhornfeldt: Do you mean in the schemas repo or is it a directory in WC? I asked in utv chat and got the answer to not use common repos

ChrilleSigerhed: Okay, I missed that conversation, sorry. A suggestion to reduce the dependency on schemas—at least partly—is to move the hash method into our new helper.

What do you think about that?

ChrilleSigerhed (approved):
Would be nice to take the oppertunity to replace HashUtility  with our new helper. Consider makeing that update before merging

PR #1141, opened 2024-10-29
K1J-449: Use limited statistics when large number of commissions

mikeyOne:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/user/UserStatisticsServiceImpl.java
```diff
@@ -71,12 +77,27 @@ public UserStatisticsDTO getUserStatistics() {
...
 
-        final var unitIds = getUnitIds(user);
+        final var careUnitIds = getCareUnitIds(user);
 
-        if (unitIds == null) {
+        if (careUnitIds.isEmpty()) {
             return null;
         }
 
+        final var useLimitedStatistics = careUnitIds.size() > maxCommissionsForStatistics;
```
ChrilleSigerhed, line 86: Should we update this variable to something that describes it better? Im thinking ```numberOfCommissionsExceeded```

ChrilleSigerhed (approved):
Nice

web/src/main/java/se/inera/intyg/webcert/web/service/facade/user/UserStatisticsServiceImpl.java
mikeyOne: Or perhaps maxCommissionsExceeded

PR #1143, opened 2024-11-01
K1J-480: Do not filter 'ArendeListItem' that has been returned from the repository as they are already correctly filtered. The removed code incorrectly filtered out answers from FK that had to be handled. Or already handled questions when querying all questions whether or not they are handled.

mww-sopra-steria:


ChrilleSigerhed (approved):
Great job!

PR #1144, opened 2024-11-01
K1J-434: When signing a certificate, that replaces/complements/extends another certificate, a HANFRFM event should be published if the other certificate contains complement questions.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #1145, opened 2024-11-01
K1J-451: Fix flaky tests that crashes if day of month is a single digit day when creating a coordination number

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #1146, opened 2024-11-04
K1J-468: Updated confirmation modal for DB

mhornfeldt:


web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificateTypesFacadeServiceImpl.java
```diff
@@ -93,6 +96,19 @@ private CertificateTypeInfoDTO addAdditionalResourceLinks(CertificateTypeInfoDTO
         return intygModule;
     }
 
+    private CertificateTypeInfoDTO addConfirmationModal(CertificateTypeInfoDTO intygModule, Personnummer patientId) {
+        final var provider = ConfirmationModalProviderResolver.get(intygModule.getId(), CertificateStatus.UNSIGNED, "NORMAL", true);
+        final var latestCertificateVersion = intygTextsService.getLatestVersion(intygModule.getId());
+        final var patient = patientDetailsResolver.resolvePatient(patientId, intygModule.getId(), latestCertificateVersion);
+        if (provider != null) {
```
ChrilleSigerhed, line 103: Should we move this if-statement so that its done right after ```final var provider = ConfirmationModalProviderResolver.get(intygModule.getId(), CertificateStatus.UNSIGNED, "NORMAL", true);```?

Should not affect performance, but seems uneccessary to ask PU for patient information if we dont need it.

Something like this:

```java
final var provider = ConfirmationModalProviderResolver.get(intygModule.getId(), CertificateStatus.UNSIGNED, "NORMAL", true);
if (provider != null) {
     final var latestCertificateVersion = intygTextsService.getLatestVersion(intygModule.getId());
     final var patient = patientDetailsResolver.resolvePatient(patientId, intygModule.getId(), latestCertificateVersion);
     intygModule.setConfirmationModal(
                provider.create(patient.getFullstandigtNamn(), patientId.getPersonnummerWithDash(), "NORMAL")
            );
}

return intygModule;
```

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/ResourceLinkFactory.java
```diff
@@ -25,6 +25,10 @@
 
 public class ResourceLinkFactory {
 
+    private ResourceLinkFactory() {
+
```
ChrilleSigerhed, line 29: Maybe add ``` throw new IllegalStateException("Utility class");```

web/src/main/java/se/inera/intyg/webcert/web/service/facade/modal/confirmation/ConfirmationModalProviderResolver.java
```diff
@@ -0,0 +1,49 @@
...
+
+package se.inera.intyg.webcert.web.service.facade.modal.confirmation;
+
+import se.inera.intyg.common.db.support.DbModuleEntryPoint;
+import se.inera.intyg.common.support.facade.model.CertificateStatus;
+
+public class ConfirmationModalProviderResolver {
+
+    private ConfirmationModalProviderResolver() {
+
```
ChrilleSigerhed, line 28: Maybe should add throw statement as already mentioned?

web/src/main/java/se/inera/intyg/webcert/web/service/facade/modal/confirmation/ConfirmationModalProviderResolver.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public static ConfirmationModalProvider get(String type, CertificateStatus status, String origin, boolean isCreatedFromList) {
+        final var isIntegratedOrigin = origin.equals("DJUPINTEGRATION");
+
+        if (!isIntegratedOrigin && !isCreatedFromList) {
```
ChrilleSigerhed, line 34: The usage of neglects is confusing, but I think I understand the purpose of this if-statement. To my understanding its so that we can handle both integrated and "not integrated" user cases. Does it make sense to instead separate the normal/integrated by different methods?

If its not possible to separate, maybe a describing method name would clarify the purpose of the if-statement?

Do we also need to consider revision? Otherwise this modal will be displayed for the user every time they open a DB, maybe its like that already? Just checking.

ChrilleSigerhed (changes requested):
Left some comments to consider. I dont feel like I fully understand the whole flow, so please contact me so that we can clarify it! :)

web/src/main/java/se/inera/intyg/webcert/web/service/facade/modal/confirmation/ConfirmationModalProviderResolver.java
mhornfeldt: Integrated -> The modal is open everytime the draft is opened
Not integrated -> The modal is opened when the draft is created from the list (Modalen visas endast en gång. Har en användare bockat i att kontroll är gjord visas inte modalen igen i fristående Webcert)

Revision is not relevant since we don't have a requirement saying do apply some requirement before the draft has been edited. 

I can rename the neglect but I don't think it's confusing. It states if the origin is not integrated and the draft is not created from list then apply rule.

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/certificatefunctions/ResourceLinkFactory.java
mhornfeldt: I don't think it's necessary since the constructor is private but I can add it

web/src/main/java/se/inera/intyg/webcert/web/service/facade/impl/GetCertificateTypesFacadeServiceImpl.java
mhornfeldt: Sure

web/src/main/java/se/inera/intyg/webcert/web/service/facade/modal/confirmation/ConfirmationModalProviderResolver.java
mhornfeldt: https://inera.atlassian.net/wiki/spaces/IT/pages/9075408/D3.1.2.+Skatteverkets+och+Socialstyrelsen+intyg

ChrilleSigerhed (approved):
Awsome, I think your update made a difference!

PR #1147, opened 2024-11-06
K1J-488: If patient is protected then filter out certificates of certificate types that user has no access to instead of throwing exception. This is because the user can have access to some certificate types e.g. nurse has access to some but not all.

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice fix!

PR #1165, opened 2024-12-19
K1J-529: Logic to handle an inactive certificate type. Inactivating a certificate type is controlled through feature toggling and an inactive type have limited actions available to it.

mww-sopra-steria:


ChrilleSigerhed (approved):
Excellent work!

Moving forward inactivating older certificates will be a peace of cake!

PR #1174, opened 2025-02-14
K1J-789: Validate that request does not generate exceeding amount of inserts.

ChrilleSigerhed:


erikagrensopra (approved):
Nice work!
Remember to add the _max.allowed.notification.send_ to the intyg-env repo before merging :)

PR #1177, opened 2025-02-17
K1J-161: Added aggregator for get intyg admin info

mhornfeldt:


mww-sopra-steria (changes requested):
Good job! I think it looks good but please check the comments. I think it would be good to deal with them even if everything will work well anyways.

A general comment to consider for the future. Now the convert methods do both fetching from WC database and calling CS. I think the code becomes more easy to read and understand if the convert methods just converts data provided. In this case we would fetch the certificate, questions, relating certificates and events and later calling the converter to transform the data as we wish.

mhornfeldt:
> Good job! I think it looks good but please check the comments. I think it would be good to deal with them even if everything will work well anyways.
> 
> A general comment to consider for the future. Now the convert methods do both fetching from WC database and calling CS. I think the code becomes more easy to read and understand if the convert methods just converts data provided. In this case we would fetch the certificate, questions, relating certificates and events and later calling the converter to transform the data as we wish. I'm not expecting you to change this and sometimes we have little options when working with legacy code, but just wanted to share the thought.

I'll think about your comment in the future, I agree with your opinion! Fixed the other changes.

mww-sopra-steria (approved):

PR #1188, opened 2025-03-10
K1J-849: Updated body of mail for questions and answers with employee name and unit id.

ChrilleSigerhed:


mww-sopra-steria (changes requested):
A few things to look into just to make sure that we handle name correct.

There is logic in ListDecoratorImpl.getNameByHsaId(...) that does similar thing that we do here. Look at it and see if there is something to consider.

Second. I understand that extracting it to a separate service simplifies testing a lot. It is now named and placed in the "mail"-package but is really not very unique to mail notification. 

Third. Reusability will increase if we build functions (when possible) that has no side-effects. For example: If we had a "service" (or maybe more as a toolkit or similar for employees) it can take the data and produce the concatenated name. Then it can be reused in places where the data has already been fetched. And it simplifies testing as you don't have to mock fetching the data.

If the change was in legacy code that probably will be removed/replaced, I would think putting too much thought in this is worth it. But now when we will definitely live with this for the duration of Webcert I think it would be good to think it through.

By the way ArendeConverter seems to be doing something similar as this code and the ListDecoratorImpl

ChrilleSigerhed:
> A few things to look into just to make sure that we handle name correct.
> 
> There is logic in ListDecoratorImpl.getNameByHsaId(...) that does similar thing that we do here. Look at it and see if there is something to consider.
> 
> Second. I understand that extracting it to a separate service simplifies testing a lot. It is now named and placed in the "mail"-package but is really not very unique to mail notification.
> 
> Third. Reusability will increase if we build functions (when possible) that has no side-effects. For example: If we had a "service" (or maybe more as a toolkit or similar for employees) it can take the data and produce the concatenated name. Then it can be reused in places where the data has already been fetched. And it simplifies testing as you don't have to mock fetching the data.
> 
> If the change was in legacy code that probably will be removed/replaced, I would think putting too much thought in this is worth it. But now when we will definitely live with this for the duration of Webcert I think it would be good to think it through.
> 
> By the way ArendeConverter seems to be doing something similar as this code and the ListDecoratorImpl

Made some changes to handle users without middleAndSurname or givenName, inspired from ArandeConverter. Also moved the logic into an employee package.

Not sure I understand the requested changes from the third section. Lets discuss it if any further changes are required.

ChrilleSigerhed:
Made changes as we discussed.

mww-sopra-steria (approved):
Nice job! Great to see that we are getting rid of some technical debt as well!

PR #1223, opened 2025-10-01
K1J-1432: Added publish event messages for delete, create from template, renew, replace, print, revoke, ready for sign

mhornfeldt:


mww-sopra-steria (approved):
Looks good! I can't think of more events that are just related to certificates. I ran some local tests and everything seems to be working fine.

PR #1224, opened 2025-10-03
K1J-1447: Fix the monitor-logging of revoke certificate when revoking certificates handles by certificate-service

mww-sopra-steria:


ChrilleSigerhed (approved):
Nice!

PR #1314, opened 2026-04-10
K1J-1524: Update configuration for JPA, Redis &  Actuator to use spring boot convention - Fix errors with RemoteIpValve lost from tomcat-jetty configuration

ChrilleSigerhed:


mww-sopra-steria (approved):
Looking good!

PR #1317, opened 2026-05-29
K1J-2145: Validate messages backend for case communication

jacobalmqvistsopra:


web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/util/EncodingValidatorBase.java
```diff
@@ -0,0 +1,61 @@
...
+
+public abstract class EncodingValidatorBase {
+
+  private static final Set<Character> ALLOWED_CHAR_SET = Set.of(new Character[] {'\n', '\r', '\t'});
+  private static final CharsetEncoder ISO_8859_1_ENCODER = StandardCharsets.ISO_8859_1.newEncoder();
+
+  protected boolean isValid(
+      String stringToCheck, ConstraintValidatorContext context, String customErrorMessage) {
+    if (stringToCheck == null || stringToCheck.isEmpty()) {
+      return false;
```
ChrilleSigerhed: Update to true

web/src/main/java/se/inera/intyg/webcert/web/web/handlers/WebcertRestExceptionHandlerAdvice.java
```diff
@@ -74,4 +78,17 @@ public ResponseEntity<WebcertRestExceptionResponse> handleRuntimeException(Runti
...
   }
+
+  @ExceptionHandler(MethodArgumentNotValidException.class)
+  public ResponseEntity<ValidationErrorResponseDTO> handleMethodArgumentNotValid(
+      MethodArgumentNotValidException ex) {
+
+    List<String> errors =
+        ex.getBindingResult().getFieldErrors().stream().map(FieldError::getDefaultMessage).toList();
+
+    return ResponseEntity.status(HttpStatus.BAD_REQUEST)
```
ChrilleSigerhed: When this type of error occurs I think its important that we can analyze our logs and find out what the cause was. Right now im getting these logs:
```
CLIENT_ERROR Received error from client with errorId '0da2e150-ffd8-43bb-8219-77c6c242669d' for certificate '9398d9eb-922e-4e2a-b13a-744f2038547c' with error code 'UNKNOWN_INTERNAL_PROBLEM', message '400 - Bad Request - POST /api/question/0cbdeef4-a504-4c74-bc44-f437f419f25a' and stacktrace 'NO_STACK_TRACE'
```
I think it will be troublesome for us to determine the cause of this ```Bad Request``` - and I suggest we make a change so that this becomes easier for us.

If we update the respone to the following:

```java
    return ResponseEntity.badRequest()
        .body(new WebcertRestExceptionResponse(
                WebCertServiceErrorCodeEnum.INTERNAL_PROBLEM, String.join("", errors)));
```

We get the following log with a error message that makes it easier for us to determine what went wrong in this scenario.
```
CLIENT_ERROR Received error from client with errorId '4b520bba-edb1-4482-b94e-0890c11eccbd' for certificate '9398d9eb-922e-4e2a-b13a-744f2038547c' with error code 'INTERNAL_PROBLEM', message 'question.message contains illegal characters: ›∂ - POST /api/question/366fb0b0-5d47-4327-92bd-ade05e8d1779' and stacktrace 'NO_STACK_TRACE'
```

ChrilleSigerhed (changes requested):
Well done, im impressed in the way you tackled this issue and I really learned something reviewing this PR. The solution is clean and effective. I added some comments to consider regarding the response that I think would be nice to add for better traceability.

It would also be nice if you added some tests to ```EncodingValidatorQuestion``` and ```EncodingValidatorString``` so that the logic is verified.

web/src/main/java/se/inera/intyg/webcert/web/web/handlers/WebcertRestExceptionHandlerAdvice.java
jacobalmqvistsopra: Great suggestion, thanks!

web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/util/EncodingValidatorQuestion.java
```diff
@@ -0,0 +1,36 @@
...
+
+import jakarta.validation.ConstraintValidator;
+import jakarta.validation.ConstraintValidatorContext;
+import se.inera.intyg.common.support.facade.model.question.Question;
+
+public class EncodingValidatorQuestion extends EncodingValidatorBase
+    implements ConstraintValidator<EncodingConstraint, Question> {
+
+  @Override
+  public boolean isValid(Question value, ConstraintValidatorContext context) {
```
ChrilleSigerhed, line 29: I think it would be a good idea to add some tests to this class and ```EncodingValidatorString``` before merging.

ChrilleSigerhed (approved):
Looks good, well done.

web/src/main/java/se/inera/intyg/webcert/web/web/controller/facade/util/EncodingValidatorQuestion.java
jacobalmqvistsopra: Yes, will add.

PR #1327, opened 2026-06-09
K1J-2196: use event timestamp when resending status updates

LucasLundstrom:


sopradanieljoensson (approved):

mww-sopra-steria (changes requested):
The implementation looks good and solved the most urgent issue. But the original implementation to handle resends has another problem that I would suggest that we should deal with straight away. It should be a minor change as well.

The PublishCertificateStatusUpdateService.resend(...) receives a NotificationRedelivery. This object contains a property that is called message (byte[]). If a message exists it should not create a new one but instead just use it. 

The notificationRedeliveryService.resend(...) takes a message as its third argument which is the message to publish. If a message exists that message should just be published directly.

mww-sopra-steria (approved):
Looks good! Nice job!

PR #1334, opened 2026-07-06
K1J-2001: Update so that subscription related integration uses RestClient. Handle updated file encoding from BOM.

ChrilleSigerhed:


integration/servicenow-integration/src/main/java/se/inera/intyg/webcert/integration/servicenow/config/ServiceNowIntegrationRestConfig.java
```diff
@@ -21,28 +21,26 @@
...
-  private int connectionTimeout;
-
   @Value("${servicenow.read.timeout}")
   private int readTimeout;
 
-  private static final String SUBSCRIPTION_SERVICE_REST_TEMPLATE = "serviceNowRestTemplate";
+  private static final String SUBSCRIPTION_REST_CLIENT = "serviceNowRestClient";
+
+  @Bean(SUBSCRIPTION_REST_CLIENT)
+  RestClient restClient() {
```
sopradanieljoensson: Maybe we can make use of auto-config here? I saw that you can use a [ClientHttpRequestFactoryBuilderCustomizer](https://docs.spring.io/spring-boot/api/java/org/springframework/boot/http/client/autoconfigure/ClientHttpRequestFactoryBuilderCustomizer.html#) to customize connection timeouts. Then the rest can be handled by auto-config. If the bean qualifier is not needed, then all of this could be removed.

integration/servicenow-integration/build.gradle
```diff
@@ -8,10 +8,13 @@ dependencies {
     implementation "jakarta.ws.rs:jakarta.ws.rs-api"
     implementation "org.apache.cxf:cxf-rt-frontend-jaxrs"
     implementation "org.springframework:spring-context-support"
+    implementation "org.apache.httpcomponents.client5:httpclient5"
     implementation "org.springframework:spring-web"
 
     annotationProcessor 'org.projectlombok:lombok'
 
+    testImplementation "org.hamcrest:hamcrest-core"
```
sopradanieljoensson: Maybe we can use the `spring-boot-starter-test` here or would that bring in too much stuff that will not be used?  [https://docs.spring.io/spring-boot/reference/testing/index.html](https://docs.spring.io/spring-boot/reference/testing/index.html)

sopradanieljoensson (approved):
Looks good, left two comments that can be looked into that would make more use of auto-config and starters.

integration/servicenow-integration/build.gradle
ChrilleSigerhed: Good idea, I added it and removed the rest since they were included in spring boot starter.

PR #1335, opened 2026-07-06
K1J-2001: Handle sonarqube warnings introduced from java 25 lift.

ChrilleSigerhed:


web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/GetUnitNotificationConfig.java
```diff
@@ -51,7 +51,10 @@ public List<RegionNotificationConfig> get() {
             "Integrated Unit Notification was loaded with configuration: {}",
             integratedUnitNotificationConfig);
       } catch (FileNotFoundException e) {
-        log.warn("File not found: {}. Returning empty configuration.", unitNotificationConfigPath);
+        log.warn(
+            "File not found: %s. Returning empty configuration."
+                .formatted(unitNotificationConfigPath),
+            e);
```
LucasLundstrom: In classes using @Slf4j you can keep the {} syntax for the message and just comma separate the exception.

web/src/main/java/se/inera/intyg/webcert/web/service/user/dto/WebCertUser.java
```diff
@@ -131,19 +129,10 @@ public boolean isUnitInactive() {
     return parameters != null && parameters.isInactiveUnit();
   }
 
-  @Override
   public String getName() {
     return "";
   }
```
LucasLundstrom, line 135: How does the relyingPartyRegistrationId get resolved now?

ChrilleSigerhed: In spring security 7 it moved to ```Saml2AssertionAuthentication```. In ```WebSecurityConfig``` we fetch it like this (already merged)

```java
 private String getRegistrationId(Saml2Authentication saml2Authentication) {
    return ((Saml2AssertionAuthentication) saml2Authentication).getRelyingPartyRegistrationId();
  }
```

web/src/main/java/se/inera/intyg/webcert/web/csintegration/certificate/GetUnitNotificationConfig.java
ChrilleSigerhed: Are you sure? I was expecting a throwable parameter as third argument if that was the case?

Image

LucasLundstrom (approved):

PR #1337, opened 2026-07-10
K1J-2262: Remove field injection via Autowired and use constructor injection

sopradanieljoensson:
Production code has had warnings mentioning not to use `@Autowired` and instead use constructor injection. This PR removes the usage of autowired and instead use RequiredArgsConstructor for constructor injection.


ChrilleSigerhed (approved):
Well done!

